PR #4, opened 2026-04-28
WIP: Add external MCPs, plugins framework, and Agent Connect

PrielHaz:
## Summary

PR walkthrough video in: https://drive.google.com/drive/folders/1CXwjkN4Q_LmWijSVbPMxfL3-KCiedmzg?usp=sharing

Adds four major pieces of functionality to the store:

1. **External MCPs** — import external MCP servers (stdio / SSE / streamable-HTTP) and use their tools as first-class store primitives. Verified with context7. To support certain mcps we should add option to run tools locally and not in container.
2. **Plugins framework + Runspace plugin** — extension system with an optional plugin for running Claude Code agents securely in Docker (agentic anthropic skill export to hierarchic directory using skill-creator skill, Store Agent chat automatically access store using the store mcp)
3. **Agent Connect tab** — UI to connect external agents to the store, exposing a curated (~20-tool) MCP alongside the existing full MCP
4. **Dependent-safe tool updates** — tools can now be safely updated/deleted when no other tool depends on them (call them); health check propagates drift across the store
5. **skill name compatibility with anthropic** — skill name validation to be consistent with anthropic and to prevent other issues

## 1. External MCPs (headline feature)

`ExternalMCPManager` owns one long-lived `ClientSession` per imported server across all three transports (`stdio`, `sse`, streamable-HTTP). On startup and on each registration it reconciles remote `list_tools()` into store primitives (`packaging_format="mcp"`, name-prefixed `<server>__<tool>`) and runs a boot-time drift sweep so dependents are marked broken when an upstream tool disappears or changes shape. Removing an external MCP preserves composites that depended on it (moved to `state=broken`, `broken_reason=dep_missing:...`) rather than deleting them.

`normalize_mcp_input` accepts five wire shapes: Claude-Desktop's `{"mcpServers": {...}}`, bare dicts, single entries, lists, and `{"source_url": "..."}` fetch. Secrets are stored plaintext and redacted on GET and on skill export.

Tool calls on external-MCP primitives go through the pooled session (fast path in `FileExecutor`, no Docker init).

Key files: `modules/external_mcp_manager.py`, `fast_api/external_mcps_api.py`, `tests/test_external_mcp.py`, `ui/src/pages/ExternalMCPsPage.tsx`, `ui/src/pages/ExternalMCPDetailPage.tsx`

## 2. Plugins framework + Runspace plugin

Protocol-based plugin system: each plugin registers a `UIManifest`, FastAPI routers, and startup/shutdown hooks. Activation lives in `$SBS_BASE_DIR/plugins/enabled.json`; admin endpoints `/plugins`, `/plugins/{id}/enable|disable` manage state. Main code iterates `active_plugins()` and never branches on a specific id.

Runspace is the only shipped plugin — optional, off by default. It talks to an external `runspace-srv` daemon to run Claude Code agents in Docker for:
- **Agentic skill export** — agent iterates and produces an Anthropic-skill bundle
- **Store Agent** — chat interface that uses the store MCP to perform operations

Endpoints previously in `skills_api.py` (`/skills/{name}/export-agent-request`, `/ai-settings`, `/agent/store-request`) were migrated to the plugin, with automatic migration of legacy `ai_features/settings.json`.

Key files: `plugins/base.py`, `plugins/__init__.py`, `plugins/runspace/*`, `ui/src/plugins/*`, `ui/src/plugins/runspace/*`

## 3. Agent Connect tab + curated Agent MCP

A second MCP server runs on its own port (`SBS_AGENT_MCP_PORT`, default 9999) with a hand-picked ~20-tool set (list/create/update/execute tools, skill + snippet + VMCP + external-MCP management, `bulk_add_tools_from_mcps`) — intentionally smaller than the auto-generated full MCP at `/control_sse` (renamed `full_mcp`) that mirrors ~40 FastAPI endpoints. The curated MCP is the better default for LLMs that can't handle the full surface area.

The Agent Connect UI shows both URLs, copy-ready `claude mcp add` commands (project/user scope toggle, curated/full toggle), and runs a live connection test.

Key files: `fast_api/agent_mcp_server.py` (new, 796 lines), `ui/src/pages/AgentConnectPage.tsx` (new, 475 lines), `fast_api/server.py` (wires `full_mcp` + `agent_mcp`), `main.py` (prints connect commands on startup)

## 4. Dependent-safe tool mutations

`tools_api.py` refuses destructive mutations (delete, interface-changing update, module swap) when another tool lists the target in its `dependencies`, returning HTTP 409 with a structured `tool_has_dependents` payload. Every mutation populates `mcp_dependencies` and `dependency_hashes` on the manifest and runs `check_all_tools_health` afterward so drift propagates across the store automatically.

New: `PUT /tools/{name}/module` (module-source updates), `PUT /tools/{name}/bundled-with-mcps` (bulk-add flag).

Health module introduces a broken-reason taxonomy: `dep_missing`, `dep_schema_changed`, `dep_code_changed`, `dep_broken`, `server_unavailable`. Broken tools surface with banners in Tool / Skill detail pages.

Key files: `fast_api/tools_api.py`, `modules/tool_health.py` (new), `schemas/tool_schema.py` (new fields + `ManifestState.BROKEN`)

## Other notable changes

- **Skill ↔ MCP bundling**: `POST /skills/{name}/bulk-add-tools-from-mcps` with three modes (`primitives`, `related`, `bundled_related`). Importer/exporter round-trip a sidecar `mcp-servers.json` inside the skill ZIP; env/header secrets are redacted to `${NAME}` placeholders on export and auto-registered on import.
- **VMCP external-MCP support**: `modules/vmcp_server.py` synthesizes MCP-primitive stubs so VMCP tool-execute routes external-MCP primitives through the pooled session.
- **Config plumbing** in `tools/configure.py`: `get_external_mcps_directory`, `get_plugins_directory`, `get_plugin_directory`, `get_ai_features_directory`. `SBSettings` gains `agent_mcp_port`.
- **UI additions**: bulk-add-from-MCPs section in Create Skill modal, broken-state banners, MCP-dependency views on Tool/Skill detail, MCP tags on tools list, dynamic plugin nav, slot-based action injection (e.g. "Export with Agent").
- **Global CSS**: ~60 lines added — worth a visual skim.
- **Fix in `check-multiarch.sh`**: removed a blank line before the shebang so the script is correctly recognized as bash on all kernels.

### DIFF
--- a/skillberry-common/scripts/check-multiarch.sh
+++ b/skillberry-common/scripts/check-multiarch.sh
@@ -1,4 +1,3 @@
-
 #!/usr/bin/env bash
 
 set -u
--- a/src/skillberry_store/fast_api/admin_api.py
+++ b/src/skillberry_store/fast_api/admin_api.py
@@ -219,6 +219,28 @@ def register_admin_api(app: FastAPI, tags: str = "admin"):
             "vector_indexes_reset": vector_indexes_reset,
         }
 
+    @app.get("/admin/server-info", tags=[tags])
+    def get_server_info():
+        """Get server connection information for MCP clients.
+
+        Returns:
+            dict: Server host, port, MCP endpoint URLs, and API docs URL.
+        """
+        settings = app.settings
+        host = settings.display_host
+        port = settings.sbs_port
+        agent_mcp_port = settings.agent_mcp_port
+
+        agent_mcp_running = hasattr(app, 'agent_mcp') and app.agent_mcp is not None
+        return {
+            "host": host,
+            "port": port,
+            "agent_mcp_port": agent_mcp_port,
+            "agent_mcp_url": f"http://{host}:{agent_mcp_port}/sse" if agent_mcp_running else None,
+            "control_mcp_url": f"http://{host}:{port}/control_sse",
+            "api_docs": f"http://{host}:{port}/docs",
+        }
+
     @app.get("/health", tags=[tags])
     def health_check():
         """Health check endpoint.
@@ -227,6 +249,95 @@ def register_admin_api(app: FastAPI, tags: str = "admin"):
             dict: Health status of the service.
         """
         return {"status": "healthy"}
+
+    @app.get("/plugins", tags=[tags])
+    def list_plugins():
+        """List all installed plugins with their current enabled state and UI manifests."""
+        from skillberry_store.plugins import installed_plugins, read_enabled
+
+        enabled_ids = set(read_enabled())
+        items = []
+        for p in installed_plugins():
+            is_enabled = p.id in enabled_ids
+            items.append({
+                "id": p.id,
+                "name": p.name,
+                "description": p.description,
+                "enabled": is_enabled,
+                "requires_restart": getattr(p, "requires_restart", False),
+                "ui_manifest": p.ui_manifest.to_dict() if is_enabled else None,
+            })
+        return {"plugins": items}
+
+    @app.post("/plugins/{plugin_id}/enable", tags=[tags])
+    def enable_plugin(plugin_id: str):
+        """Enable a plugin.
+
+        For plugins with `requires_restart=False`, routes are mounted live.
+        Otherwise `enabled.json` is updated but the response signals that a
+        store restart is required for the change to take effect.
+        """
+        from skillberry_store.plugins import (
+            get_plugin,
+            is_enabled,
+            mount_plugin,
+            set_enabled,
+        )
+
+        plugin = get_plugin(plugin_id)
+        if plugin is None:
+            raise HTTPException(status_code=404, detail=f"Plugin '{plugin_id}' not installed")
+
+        already = is_enabled(plugin_id)
+        set_enabled(plugin_id, True)
+
+        if plugin.requires_restart:
+            return {"plugin_id": plugin_id, "enabled": True, "restart_required": True}
+
+        if not already:
+            try:
+                if hasattr(plugin, "bind_app"):
+                    plugin.bind_app(app)
+                mount_plugin(app, plugin)
+            except Exception as e:
+                logger.error(f"Failed to mount plugin '{plugin_id}': {e}")
+                raise HTTPException(status_code=500, detail=f"Mount failed: {e}")
+
+        return {"plugin_id": plugin_id, "enabled": True, "restart_required": False}
+
+    @app.post("/plugins/{plugin_id}/disable", tags=[tags])
+    def disable_plugin(plugin_id: str):
+        """Disable a plugin.
+
+        For plugins with `requires_restart=False`, routes are removed live.
+        Otherwise `enabled.json` is updated but the response signals that a
+        store restart is required for the change to take effect.
+        """
+        from skillberry_store.plugins import (
+            get_plugin,
+            is_enabled,
+            set_enabled,
+            unmount_plugin,
+        )
+
+        plugin = get_plugin(plugin_id)
+        if plugin is None:
+            raise HTTPException(status_code=404, detail=f"Plugin '{plugin_id}' not installed")
+
+        was_enabled = is_enabled(plugin_id)
+        set_enabled(plugin_id, False)
+
+        if plugin.requires_restart:
+            return {"plugin_id": plugin_id, "enabled": False, "restart_required": True}
+
+        if was_enabled:
+            try:
+                unmount_plugin(app, plugin)
+            except Exception as e:
+                logger.error(f"Failed to unmount plugin '{plugin_id}': {e}")
+                raise HTTPException(status_code=500, detail=f"Unmount failed: {e}")
+
+        return {"plugin_id": plugin_id, "enabled": False, "restart_required": False}
 
     @app.get("/health/ready", tags=[tags])
     def readiness_check():
--- a/src/skillberry_store/fast_api/skills_api.py
+++ b/src/skillberry_store/fast_api/skills_api.py
@@ -2,10 +2,11 @@
 
 import json
 import logging
+import os
 import uuid
 from datetime import datetime, timezone
-from typing import Optional, Annotated
-from fastapi import FastAPI, HTTPException, Query, UploadFile, File, Form
+from typing import Any, Dict, List, Optional, Annotated
+from fastapi import Body, FastAPI, HTTPException, Query, Request, UploadFile, File, Form
 from fastapi.responses import Response
 from prometheus_client import Counter
 
@@ -21,6 +22,7 @@ from skillberry_store.tools.configure import (
     is_auto_detect_dependencies_enabled,
 )
 from skillberry_store.fast_api.search_filters import apply_search_filters
+from skillberry_store.fast_api.skill_export_utils import get_skill_export_data
 
 logger = logging.getLogger(__name__)
 
@@ -413,6 +415,7 @@ def register_skills_api(
 
     @app.post("/skills/import-anthropic", tags=[tags])
     async def import_anthropic_skill(
+        request: Request,
         source_type: str = Form(...),
         github_url: Optional[str] = Form(None),
         zip_file: Optional[UploadFile] = File(None),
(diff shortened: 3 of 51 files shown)


skillberry-common/scripts/check-multiarch.sh
PrielHaz, line 1: important. its a bug that cause crushes because the ! is not first line

eranra (commented):
@PrielHaz Thanks for working on this PR and for the very nice improvment :-) 

Please split this PR into smaller PR's we are unable to review such a huge PR and approve. 

In additon --- I saw that you changed chnaged the tool_schema and added a lot of specific fields ... things like dependency_hashes, bundled_with_mcps etc ... those fields are fields that I would (a) not want to make a primary optional fileds in the schem of tools ... they are specific to the MCP case and do not need to be surfaced in the main schema fileds. 

In addition there is an already existing implementation of MCP tools in the code (look in the existing tests) ... those changes  replicate things that we alreayd have implemented. I think that there are areas that this extends but we need to crisp this ... 

The main thing is ... please breat this into simplere and smallers PR's that will allow us to review

PrielHaz:
Thanks for the feedback, makes sense. I’ll work on splitting this into smaller, more focused PRs to make review manageable.

Regarding the existing MCP tool import implementation, my understanding was that it covers a somewhat different use case. From what I saw (and from checking with Claude Code), that existing import flow supports importing one tool at a time from a running MCP server only via SSE, and now we support across all three transports (SSE, HTTP, and JSON-RPC over stdio) and import all together.
It also depends on the server already being up, so users cannot provide their existing MCP server JSON configs and have the store import and work from those definitions directly.

What I was trying to enable here is different: when the store starts, it can take the MCP JSON configurations users already have in their agents, support the different JSON formats(5 total) and all three MCP transport types(sse, http, and json-rpc with stdio), start the servers, and communicate with them directly without requiring users to manually run external server to upload tool from(also not all servers provide a url.. json-rpc with stdio not provides any url and must run some npx command to start communicating through stdio)

Part of the motivation was also integration. This is highly integrative both for users and for Kagenti, because it accepts the MCP configuration format users already have.

eranra:
> Thanks for the feedback, makes sense. I’ll work on splitting this into smaller, more focused PRs to make review manageable.

@PrielHaz thanks --- also FYI @aviweit and myself spent time on the current implementation of the MCP tools in the store ... I asked @aviweit to move all the MCP specific details into the "extra" field ... so this can align nicely with the changes that you did with the tool schema  - @aviweit @PrielHaz please discuss.

Last on that @PrielHaz I think that a good split into smaller PR's each focused on specific extention will make this much more easy to consume and review. 

THX again ... TTYT.

PrielHaz:
Thanks, I’ll split it into smaller PRs.

Regarding the MCP metadata, I’m not sure fields like whether a tool is imported from MCP or has MCP dependencies should live under "extra", since I see them as important tool-level fields, even if they’re optional and mainly relevant for MCP-imported tools or composites that recursively depend on MCP tools.

These fields are also used by the skill creation flow, for example to enable adding all MCP-related tools into a skill and support related functionality.

What do you think is the right place for these fields?

src/skillberry_store/fast_api/tools_api.py
```diff
@@ -466,6 +570,37 @@ def delete_tool(name: str) -> Dict:
                 status_code=500, detail=f"Error deleting tool: {str(e)}"
             )
 
+    @app.put("/tools/{name}/bundled-with-mcps", tags=[tags])
```
aviweit, line 573: I am not sure if mcp specific endpoints should be inside tools_api.py.

src/skillberry_store/fast_api/tools_api.py
aviweit, line 250: Is module relevant when adding an MCP tool? I think the code itself is in the MCP server already.

aviweit, line 296: I think it is only relevant for none mcp.

src/skillberry_store/plugins/__init__.py
aviweit, line 1: I think the plugin framework along with runspace can go under a separate PR.

src/skillberry_store/schemas/manifest_schema.py
```diff
@@ -14,6 +14,7 @@ class ManifestState(str, Enum):
     NEW = "new"
     CHECKED = "checked"
     APPROVED = "approved"
+    BROKEN = "broken"
```
aviweit, line 17: I wonder if "error" can be used instead?

src/skillberry_store/fast_api/skills_api.py
```diff
@@ -598,103 +639,154 @@ async def import_anthropic_skill(
...
+            touches_requested = (tool_mcp in mcp_set) or bool(tool_deps & mcp_set)
+            if not touches_requested:
+                continue
+
+            if mode == "primitives" and tool_mcp not in mcp_set:
+                continue  # composites excluded
+            if mode == "bundled_related" and d.get("bundled_with_mcps") is False:
+                skipped_unbundled.append(d.get("name") or fname[:-5])
+                continue
+            if d.get("state") == "broken":
```
aviweit, line 719: Please use the enumeration value (BROKEN)

src/skillberry_store/modules/tool_health.py
aviweit, line 1: I think it can go into a separated PR.

src/skillberry_store/fast_api/admin_api.py
```diff
@@ -228,6 +250,95 @@ def health_check():
         """
         return {"status": "healthy"}
 
+    @app.get("/plugins", tags=[tags])
```
aviweit, line 253: Maybe this and other plugin endpoints should go into the ‘plugin enablement’ PR?

src/skillberry_store/schemas/tool_schema.py
```diff
@@ -80,7 +80,36 @@ class ToolSchema(ManifestSchema):
...
+    )
+    mcp_server: Optional[str] = Field(
+        default=None,
+        description="Name of the external MCP server this tool was imported from; None for user-authored tools."
+    )
+    mcp_dependencies: List[str] = Field(
+        default_factory=list,
+        description="External MCP server names this tool requires at runtime (union across dependencies for composites)."
+    )
+    dependency_hashes: Dict[str, Dict[str, str]] = Field(
```
aviweit, line 98: It seems not to be related to MCP. If that's correct, I think it would be good if it goes into a separate PR.

src/skillberry_store/schemas/tool_schema.py
```diff
@@ -80,7 +80,36 @@ class ToolSchema(ManifestSchema):
...
+        default_factory=dict,
+        description="Map of dep_name -> {params, module, combined} sha256 hashes captured at create/update time; used by the health pass to attribute drift to schema vs code."
+    )
+    broken_reason: Optional[str] = Field(
```
aviweit, line 102: I wonder whether this could also be applied to other resources, such as VMCP servers (for example, corrupted or unreachable servers). If so, perhaps it should be considered for inclusion in the base class.

src/skillberry_store/ui/src/plugins/types.ts
aviweit, line 1: Should probably go into the "plugin" PR.

aviweit (commented):
@PrielHaz @eranra please find inline comments from my side. Thanks

src/skillberry_store/schemas/tool_schema.py
PrielHaz: The "broken" i think is in the tool level, since unlike tools we have in the store which stay consistent, MCP external servers can update the tools schemas arguments etc and therefore brake tools in the store that assume the old schema,
Then i think we should maybe mark VMCPs as broken if they have >0 broken tools

PrielHaz:
@eranra regarding the mcp fields - they are already optional and applied only to mcp imported tools.
Most of the tools people use today are from mcps so i dont think we should put all the mcp related fields under some extra field..

src/skillberry_store/schemas/tool_schema.py
PrielHaz: It's actually the main reason the field exists — the generic naming hid that.

`dependency_hashes` is the baseline used by the new health pass (`tool_health.py`) to detect when a tool's dependency has silently drifted. For local tools this is mostly a nice-to-have (deps only change on a commit). For MCP-imported primitives it's load-bearing: every time an external server re-advertises its tool list, `ExternalMCPManager` overwrites the primitive's `params` with whatever the remote schema currently is — no local edit involved. Without a recorded baseline, a Slack/GitHub/etc. server could change a primitive's required fields and every composite depending on it would keep looking "approved" until a call blew up at runtime.

How it plays with the other MCP failure modes:

| Upstream change           | Detected by                              | broken_reason                                          |
|---------------------------|------------------------------------------|--------------------------------------------------------|
| Server down / won't start | ExternalMCPManager                       | server_unavailable:<server> → composites get dep_broken:<prim> |
| Server up, schema changed | health pass via dependency_hashes        | dep_schema_changed:<prim>                              |
| Server up, tool removed   | health pass (primitive missing)          | dep_missing:<prim>                                     |

So splitting it out would mean shipping external MCP support without the thing that notices schema drift, which is the failure mode unique to external deps.

PR #21, opened 2026-04-29
Updated Docker support

erezh16:
**1. (bug fix) moved volume settings to docker-run**
**2. Removed all image push targets since buildx handles** 
**3. Fixed podman support** 
**4. Added docker-pull target**
**5. docker-run tries docker-pull first unless SBD_DEV is set**
Once the PR merges, I will update `skillberry-common` to persist fix across all projects

### DIFF
--- a/skillberry-common/.mk/docker.mk
+++ b/skillberry-common/.mk/docker.mk
@@ -4,9 +4,9 @@
 SUPPORTED_ARCHS := linux/amd64 linux/arm64
 
 # Docker registry host
-REGISTRY_HOST ?= skillberry-1.vpc.cloud9.ibm.com:8800
+REGISTRY_HOST ?= ghcr.io
 # change DOCKER_PROJECT to "skillberry" for public images
-DOCKER_PROJECT ?= skillberry-dev
+DOCKER_PROJECT ?= skillberry-ai
 # Docker repo name
 REPOSITORY_NAME = $(REGISTRY_HOST)/$(DOCKER_PROJECT)
 # Docker command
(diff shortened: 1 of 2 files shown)


skillberry-common/.mk/docker.mk
```diff
@@ -102,33 +102,58 @@ multiarch-check:
 	@echo "Verifying multi-arch container build is enabled and supports: $(SUPPORTED_ARCHS)"
 	@$(SB_COMMON_PATH)/scripts/check-multiarch.sh $(DOCKER) $(SUPPORTED_ARCHS) || exit 1
 
+# DBT - Docker Build Target: "local" for local build and "registry" for pushing the built image to registry. Default - local
```
eranra, line 113: The default should be the registry ... there is no logic to build locally ... it is useful only for debug

skillberry-common/.mk/docker.mk
```diff
@@ -102,33 +102,58 @@ multiarch-check:
...
+DBT ?= local
+
+ifeq ($(DBT),local)
```
eranra, line 116: It is very complicated and not common to go that path ... the common way is to just state the local registry as localhost:5000 or sometihng like that ... please try to remove all this complexity ... it makes the bash/Make code supper comlicated and not readable ... 

This is not good for a simple project like SKillberry ... I would ask to try and aoide all this complexity.

skillberry-common/.mk/docker.mk
```diff
@@ -171,69 +196,123 @@ docker-build: docker-check update-git-version .stamps/docker-build	## Build serv
...
-		--build-arg SERVICE_NAME="$(SERVICE_NAME)" \
-		--build-arg SERVICE_PORTS="$(SERVICE_PORTS)" \
-		--build-arg SERVICE_ENTRY_MODULE="$(SERVICE_ENTRY_MODULE)" \
-		--ssh default=$$SSH_AUTH_SOCK \
-		$(VOLUME_FLAGS) \
-		-t $(FULL_IMAGE_NAME):$(IMAGE_TAG) \
-		-t $(FULL_IMAGE_NAME):latest \
-		. || exit 1; \
-		touch .stamps/docker-build; \
+		if [ "$(DBT)" = "registry" ]; then \
```
eranra, line 214: SO much repeated code between the two options ... please try to consolidate.

skillberry-common/.mk/docker.mk
```diff
@@ -171,69 +196,123 @@ docker-build: docker-check update-git-version .stamps/docker-build	## Build serv
...
+		$(DOCKER) tag $(FULL_IMAGE_NAME):latest $(FULL_IMAGE_NAME):$(IMAGE_TAG); \
+		echo "✓ Tagged as: $(FULL_IMAGE_NAME):$(IMAGE_TAG)"; \
+		touch .stamps/docker-get; \
+	else \
+		echo "✗ Failed to pull image: $(FULL_IMAGE_NAME):latest"; \
+		exit 1; \
+	fi
+
+.PHONY: docker-get
+docker-get: .stamps/docker-get
```
eranra, line 280: Feels like over complexit and making things complicated with all the "stamps"  --- you make all of this unmainainable for anyone that will try to fix the code in the future ... please reduce the level of complexisty ... it doesn't help ... !

skillberry-common/.mk/docker.mk
```diff
@@ -171,69 +196,123 @@ docker-build: docker-check update-git-version .stamps/docker-build	## Build serv
...
+		echo "✓ Using pulled image"; \
+	else \
+		echo "⚠ Pull failed - falling back to building image locally"; \
+		$(MAKE) docker-build; \
+	fi
+endif
 
 .PHONY: docker-run
 ifeq ($(USE_LLM_SVCS),1)
-docker-run: docker-check docker-build docker-clean check-rits-watsonx-envs
```
eranra, line 214: We need to handle the check-rits-watsonx-envs --- I was not expecting that as part of the common ... but I didn't thought well enought ... we need to clean this out. 

Please open an issue. 

Also, we need to ask @osher to look on his code inside common and fix ... can you also open an issue on that  @erezh16 

^^^^

skillberry-common/.mk/docker.mk
erezh16: Correct. Hence `SBD_DEV`. `DBT` is Docker Build Target - when you actually build, which is not the default.

skillberry-common/.mk/docker.mk
erezh16: This is a podma limitation. No `buildx`. So I have to split handling of DBT cases inside.

skillberry-common/.mk/docker.mk
erezh16: stamps are a common mechanism for not repeating long operations. The phony target depends on stamp, which is a real file. The file is touched if the operation succeeds. That's all.

skillberry-common/.mk/docker.mk
erezh16: This is a usability enhancement in IBM, because using RITS ot watsonx requires their specific credentials. However, I think we can generalize this to `check-credentials` once all code uses `llm-client`. Will open an issue.

skillberry-common/.mk/docker.mk
erezh16: There is no local registry in public GH. It's GHCR - check the settings in the beginning of `docker.mk`. We can have a f2f review and I explain this code, which is not complex at all IMHO.

eranra:
@erezh16 I am approving the PR.

FYI: The current implementation feels somewhat over-engineered. It might be worth opening an issue to explore opportunities for simplification, as the existing complexity could make future maintenance challenging.

eranra (approved):

erezh16:
1. Added documentation to `docker.mk`
2. Changed root image to `python:3.11-slim` on docker hub
3. Fixed a minor bug when missing `.env` in `docker-run`
4. Added observavbility to track some changes (no behavior change)

PR #74, opened 2026-05-05
Validate skill & VMCP names against Anthropic slug format

### LINKED ISSUE #75: Enforce Anthropic-compatible slug format for skill & VMCP names
## Problem

Skill and VMCP server names created via the store are free-form strings today. Names containing spaces, uppercase letters, underscores, dots, or other punctuation pass validation at create-time but later break downstream when those names get used as:

1. **`claude mcp add <name> …` CLI positional args** — the upstream CLI rejects names outside the Anthropic Agent Skills slug format.
2. **URL path segments** — e.g. `/skills/{name}`, `/vmcp/{name}`, and skill-export URLs.
3. **SKILL.md frontmatter** — Anthropic's Agent Skills spec requires `name` to match `^[a-z0-9-]{1,64}$` (lowercase letters, digits, hyphens; 1–64 chars).

Concretely: a user can create a skill called `"My Skill"` via the UI and the store accepts it, but exporting it as an Anthropic skill or wiring it into Claude Code as an MCP later fails with an opaque error.

## Proposal

Validate skill and VMCP names against `^[a-z0-9-]{1,64}$` at create time.

- **Backend:** shared helper (`schemas/name_validation.py`) raising `HTTPException(400, detail={error: "invalid_name", …})` with a structured hint; hook into `POST /skills/` and `POST /vmcp/`.
- **Frontend:** mirror the rule in `services/api.ts` so the Create Skill modal can surface invalid names inline before submit. Backend remains authoritative.

Source of truth: Anthropic Agent Skills SKILL.md frontmatter — https://code.claude.com/docs/en/skills

## Out of scope

- External MCP server name validation, tool name validation — follow-ups, each in its own PR.
- Retroactive migration of existing skills/VMCPs with invalid names — none exist in current fixtures; if that changes, add a one-off migration.

## Fix

See PR #74.

PrielHaz:
Closes #75

## Summary

Reject skill and VMCP names that don't match Anthropic's Agent Skills slug format: `^[a-z0-9-]{1,64}$` (lowercase alpha-num + hyphens, 1–64 chars). Prevents creating entities whose names later break `claude mcp add <name>` or fail URL-segment parsing when exported / consumed as MCP.

Split out of #4 as a small, standalone PR.

## Commits (in order)

1. `chore: gitignore CLAUDE.md` — per-user Claude Code notes shouldn't be tracked.
2. `feat(schemas): add Anthropic-compatible name validation helper` — new `schemas/name_validation.py` with `STORE_NAME_PATTERN`, `validate_store_name` (raises HTTP 400), and `validate_store_name_message` (non-raising). No callers yet.
3. `feat(skills): validate skill name on create` — 5-line hook in `create_skill`.
4. `feat(vmcp): validate VMCP server name on create` — 5-line hook in `create_vmcp`.
5. `feat(ui): export STORE_NAME_PATTERN / isValidStoreName helpers` — TS mirror in `services/api.ts`.
6. `feat(ui): validate skill name inline in Create Skill modal` — field flips to `validated='error'` with a helper message as you type; submit guard matches the backend 400.

## Why duplicate the regex in TS?

The UI runs in the browser and can't import Python. The backend `validate_store_name` on `POST /skills/` remains the source of truth — the TS mirror is purely for UX (inline feedback before submit). If the two drift, the backend still rejects with 400.

## Test plan

- [ ] `POST /skills/` with `name="My Skill"` / `"foo_bar"` / uppercase → 400 `invalid_name` with structured detail
- [ ] `POST /skills/` with `name="docs-research"` → 200 as before
- [ ] `POST /vmcp/` same matrix → same behavior
- [ ] Create Skill modal: typing an invalid name paints the field red with the hint; submit is blocked with the same message; clearing the field resets validation
- [ ] Typing a valid name clears the error and submit proceeds

## Out of scope

External MCP name validation, tool name validation, Agent Connect MCP, and the broader external-MCP / plugins work from #4 — each will land as its own PR.


PrielHaz:
Addressed the Copilot review in f42a899. Summary of what changed and why:

**Validation gaps (fixed)**
- `PUT /skills/{name}` now re-validates the URL name and forces `skill.name = name`, so the detail-page edit flow can't introduce a non-slug name.
- `PUT /vmcp_servers/{name}` gets the same treatment (names are immutable after create).
- `POST /skills/import-anthropic` now slugifies the upstream SKILL.md `name` field via a new `slugify_store_name()` helper, then validates. Legacy underscore names (e.g. `test_anthropic_skill` → `test-anthropic-skill`) import cleanly; anything that still can't be made into a valid slug is rejected with the same 400 payload.

**UI error rendering (fixed)**
- `handleResponse` now handles dict-shaped FastAPI `detail` payloads (reads `.hint`, then falls back to `.error` / `.name` / JSON). No more `[object Object]` when the backend rejects a name.
- Create Skill modal: keystrokes in the name field clear the stale submit error banner; the inline validation hint is rendered via PatternFly `FormHelperText` + `HelperTextItem` and wired to the input via `aria-describedby` so screen readers announce it.
- Narrowed the misleading comment in `api.ts` — the regex is currently used for skills and VMCPs only; tools and external MCPs are explicitly out of scope for this PR.

**Tests**
- New unit tests in `src/skillberry_store/tests/test_name_validation.py` (28 cases) covering `is_valid_store_name`, `validate_store_name` (raises structured 400), and `slugify_store_name` (valid conversions + unrecoverable inputs → `None`).
- New parameterized e2e cases asserting `POST /skills/` and `POST /vmcp_servers/` reject underscore/uppercase/space/dot/empty/too-long names with `{error: "invalid_name"}` detail.
- New e2e cases asserting `PUT /skills/{name}` and `PUT /vmcp_servers/{name}` reject invalid URL names.
- Fixed `test_anthropic_api.py` assertion to match the slugified output.
- Renamed the three remaining underscore skill names in `integration/test_skills_integration.py` that would have flipped from 200→400.

**Pushed back on**
- "Renamed local variable uses a hyphen (invalid Python syntax)": verified — `python3 -c "ast.parse(...)"` on `test_skills_api.py` parses fine. All the hyphens Copilot flagged are inside string literals (names used as API values), not Python identifiers. Nothing to change.
- "Admin import/restore of older exports with underscore names will now be rejected": real trade-off, but the tightening is deliberate — the whole point of the PR is that these names must match the Anthropic slug spec going forward. Migrating legacy backups is a separate concern (e.g. a one-shot migration on admin import) that I'd rather handle in a follow-up PR than dilute the invariant here.

src/skillberry_store/schemas/name_validation.py
```diff
@@ -0,0 +1,109 @@
...
+from __future__ import annotations
+
+import re
+from typing import Any, Dict
+
+from fastapi import HTTPException
+
+# Anthropic Agent Skills naming spec.
+STORE_NAME_PATTERN = r"^[a-z0-9-]{1,64}$"
+STORE_NAME_RE = re.compile(STORE_NAME_PATTERN)
```
aviweit, line 31: I wonder whether they can be renamed to
`ANTROPIC_NAME_RE_PATTERN`
`ANTROPIC_NAME_RE`

src/skillberry_store/fast_api/skills_api.py
```diff
@@ -295,6 +303,14 @@ def update_skill(name: str, skill: SkillSchema):
         update_skill_counter.inc()
 
         try:
+            # Skill names are immutable post-creation: they are the on-disk
+            # filename, the URL segment, and the MCP server ident. Reject
+            # attempts to rename via the body, and re-validate the path `name`
+            # so we never persist an invalid slug even if one slipped in
+            # before this check existed.
+            validate_store_name(name, kind="skill")
+            skill.name = name
```
aviweit, line 312: Maybe I’m missing something, but I think the name value is currently immutable and cannot be updated.

src/skillberry_store/schemas/name_validation.py
```diff
@@ -0,0 +1,109 @@
...
+                "error": "invalid_name",
+                "kind": kind,
+                "name": name,
+                "hint": format_hint(kind),
+                "pattern": STORE_NAME_PATTERN,
+            },
+        )
+
+
+def slugify_store_name(name: Any) -> str | None:
```
aviweit, line 76: Does it make sense to renamed to `slugify_skill_name` ?

src/skillberry_store/schemas/name_validation.py
```diff
@@ -0,0 +1,109 @@
...
+def format_hint(kind: str) -> str:
+    return (
+        f"{kind} names must match the Anthropic Agent Skills format: "
+        f"lowercase letters, digits, and hyphens (-) only; 1 to 64 characters. "
+        f"No spaces, underscores, uppercase, dots, or other punctuation. "
+        f"Examples: 'docs-research', 'context7', 'pdf-to-markdown'."
+    )
+
+
+def validate_store_name(name: Any, kind: str = "name") -> None:
```
aviweit, line 47: I think it would be better to rename this method - we are not validating the store name itself.

src/skillberry_store/fast_api/skills_api.py
```diff
@@ -104,6 +108,10 @@ def create_skill(skill: Annotated[SkillSchema, Query()]):
         logger.info(f"Request to create skill: {skill.name}")
         create_skill_counter.inc()
 
+        # Skill names must be Anthropic-slug compatible — they become
+        # `claude mcp add <name>` args and URL segments.
+        validate_store_name(skill.name, kind="skill")
```
aviweit, line 113: I think that the skill maker calls this endpoint to create a skill. We need to ensure it does not break and the skill maker uses compatible skill names so that skill creation succeeds. Please double check this with @bcarmeli  and @OsherElhadad .

src/skillberry_store/fast_api/vmcp_api.py
```diff
@@ -108,6 +109,10 @@ def create_vmcp_server(vmcp: Annotated[VmcpSchema, Query()], request: Request):
         logger.info(f"Request to create vmcp server: {vmcp.name}")
         create_vmcp_counter.inc()
 
+        # VMCP names are used as `claude mcp add <name>` args and URL
+        # segments, so they must match the Anthropic slug format.
+        validate_store_name(vmcp.name, kind="VMCP server")
```
aviweit, line 114: Several skill-berry components call this endpoint (tau2 mcp agent, proxy agent, ..), we need to ensure the name they pass are compatible (i.e. in some cases uuid of the env_id is being appended).

src/skillberry_store/fast_api/vmcp_api.py
```diff
@@ -437,6 +442,13 @@ def update_vmcp_server(
         update_vmcp_counter.inc()
 
         try:
+            # VMCP server names are immutable post-creation: they are the
+            # on-disk filename, the URL segment, and the MCP server ident.
+            # Reject any rename attempt via the body and re-validate the
+            # path `name` so we never persist an invalid slug.
+            validate_store_name(name, kind="VMCP server")
+            vmcp.name = name
```
aviweit, line 450: Maybe I’m missing something, but I think the name value is currently immutable and cannot be updated.

src/skillberry_store/ui/src/services/api.ts
```diff
@@ -13,6 +13,18 @@ import type {
 
 const API_BASE = '/api';
 
+// Anthropic Agent Skills naming format — currently enforced for skill and
+// VMCP names on create/update. External MCP and tool names are out of scope
+// for this iteration, but share the same pattern so the constants are
+// exported here for future reuse.
+// Source: https://code.claude.com/docs/en/skills frontmatter.
+export const STORE_NAME_PATTERN = /^[a-z0-9-]{1,64}$/;
```
aviweit, line 21: similar comment in name_validation.py: maybe rename ?

aviweit (commented):
@PrielHaz , please find inline comments from my side.
Thanks.

erezh16:
```
External MCP name validation, tool name validation, Agent Connect MCP, and the broader external-MCP / plugins work from https://github.com/skillberry-ai/skillberry-store/pull/4 — each will land as its own PR.
```
I think ir makes more sense to add all name validation into a single PR. Saves review time and keeps the context coherent

src/skillberry_store/fast_api/skills_api.py
```diff
@@ -133,6 +134,10 @@ def create_skill(skill: Annotated[SkillSchema, Query()]):
...
```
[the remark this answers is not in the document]
erezh16: Not critical in my view. However, if you want, you can mitigate by adding a conversion utility.

erezh16:
After discussing with @bcarmeli , I now believe this validation at the store level is a bit too much. Anthropic may not be the only skill format for long because the market keeps changing. Therefore, we should not enforce its name compatibility in the store.
We can still use name validation in the following points in the workflow:

1. *Pre-generation*: we provide a `skill-maker` and/or `tool-maker` **config switch** (GUI and API) to generate Anthropic-compatible skills, tools, etc.

2. *Post-export*: we provide a **utility** that can try to convert an exported skill to Anthropic compatibility, in case of an issue such as @PrielHaz  described above. We can even integrate the validation in the export itself and offer automatic fix.

Either way, **I don't think we should merge this PR right now**. There is not enough agreement on this feature. Let's wait until @eranra comes back and discuss with him.

PrielHaz:
Closing this PR — we don't want to merge it as-is.

The underlying problem it addresses is real and stays tracked in #75: skill & VMCP names today are free-form, so a name like `"My Skill"` (spaces, uppercase, underscores, dots) is accepted at create-time but later breaks:

- `claude mcp add <name> …` — the upstream CLI rejects names outside the Anthropic Agent Skills slug format.
- URL path segments — `/skills/{name}`, `/vmcp/{name}`, skill-export URLs.
- Anthropic SKILL.md frontmatter, which requires `name` to match `^[a-z0-9-]{1,64}$`.

#75 remains open to capture this. The validation will be reworked and re-land on its own rather than through this branch.

PR #83, opened 2026-05-06
Refactor UUID lookups for skills and VMCP flows

aviweit:
# Refactor UUID lookups for skills and VMCP flows

## Summary
This PR reduces repeated full-directory scans when resolving tools and snippets by UUID.

## Changes
- add shared UUID lookup utilities in `src/skillberry_store/modules/lookup_index.py`
- refactor skills and VMCP flows to reuse prebuilt lookup maps
- add focused tests in `src/skillberry_store/modules/test_lookup_index.py`

## Why
Previously, several code paths repeatedly scanned all files while hydrating related objects. This change builds lookup maps once per operation and reuses them for O(1) UUID resolution.

## Notes
- behavior is intended to remain unchanged


src/skillberry_store/fast_api/skills_api.py
```diff
@@ -677,73 +660,51 @@ async def export_anthropic_skill(name: str):
...
+                    if not tool_dict:
+                        continue
+
+                    tools.append(tool_dict)
+                    tool_name = tool_dict["name"]
+                    lang = tool_dict.get("programming_language", "python").lower()
+                    ext = ".py" if lang == "python" else ".sh"
+                    module_filename = f"{tool_name}{ext}"
+                    try:
+                        from skillberry_store.tools.configure import (
```
eranra: I really prefere not to import  in the middle of the code if this is not totally necesary

src/skillberry_store/fast_api/vmcp_api.py
```diff
@@ -655,65 +596,34 @@ def start_vmcp_server(name: str, request: Request):
                 skills_handler = FileHandler(get_skills_directory())
                 tools_handler = FileHandler(get_tools_directory())
                 snippets_handler = FileHandler(get_snippets_directory())
+                lookup_context = build_lookup_context(
+                    skills_handler=skills_handler,
+                    tools_handler=tools_handler,
+                    snippets_handler=snippets_handler,
+                )
+
+                skill_dict = lookup_context.skills_by_uuid.get(skill_uuid)
```
eranra: Why do we do this here, and not near the tools code in line 492?

src/skillberry_store/modules/lookup_index.py
```diff
@@ -0,0 +1,96 @@
...
+
+from fastapi import HTTPException
+
+from skillberry_store.modules.file_handler import FileHandler
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class LookupContext:
```
eranra: I think that this is essentially a "cache" --- can we rename all of this to be a cache and not index? 

Also, assuming really huge scale and not midium scale - I do not like that we cache everything "in-memory" this breaks the logic of file-system (or DB) backend -- so I propose to limit this to LRU set of "skills" and not to all the skills in the system --- this way we accelerate the usage and view of skills that are really in-used but we do not load everything into memory.

eranra (commented):
I really like this --- please look on the comments

eranra (commented):
As you can see @aviweit lint is failing

src/skillberry_store/fast_api/skills_api.py
```diff
@@ -188,6 +157,10 @@ def list_skills():
         list_skills_counter.inc()
 
         try:
```
eranra, line 161: I undersatnd that this accelerate but another improvment that I am looking for is something to just return the UUIDs of the snippets and tools that belong to the skill --- not the details of the tools and snippets ... we can get the details with additional API calls for the relevant tools and skills - why are we doing this here? this is not scalable if we have a lot of skills. 

I agree that we can get all the tools and skills details for a single skill but not when we get all of the skills.

@aviweit  ^^^

src/skillberry_store/fast_api/skills_api.py
aviweit: Yes, done.

aviweit:
> As you can see @aviweit lint is failing

Fixed

src/skillberry_store/fast_api/skills_api.py
aviweit: @eranra , it makes sense. Can it be added under a separate (successive) PR?

eranra: YES ... please open an issue ... the main question is if this breakes other projects that uses the store ... we need to check that.

eranra:
@aviweit reminder : we talked about scaling to 100 skills :-)  as a basic scale goal.

src/skillberry_store/modules/lookup_index.py
aviweit: Good idea. I will look into it.

aviweit:
We need to ensure the solution is scalable for:

* Listing skills (with and without filtering)
* Searching skills (with and without filtering)

This applies to both the UI, CLI and REST (SDK).

Consider placing a script (under contrib/) that downloads and imports skills (e.g from skills.sh) into the store. It could be useful for benchmarks and demonstration purpose.

src/skillberry_store/fast_api/skills_api.py
aviweit: > YES ... please open an issue ... the main question is if this breakes other projects that uses the store ... we need to check that.

Yes, the below will be validated:

**Tool Maker Validation**
   - Run tool maker against the store

**Benchmark Testing**
   - Run benchmarks repository tests:
     - **Tau2 MCP Agent**: Test MCP protocol compatibility
     - **Proxy Agent**: Test agent proxy functionality

**Skill Maker Validation**
   - Run skill maker against the store

PR #154, opened 2026-05-28
Delete Race Fix

erezh16:
- **1. Fixed race condition on DELETE: cache should be updated before object removal 2. Escalated the warning in the log to error + HTTPException 500**
- **Lint fix**

### DIFF
--- a/src/skillberry_store/fast_api/skills_api.py
+++ b/src/skillberry_store/fast_api/skills_api.py
@@ -277,11 +277,7 @@ def register_skills_api(
             except Exception as e:
                 logger.warning(f"Could not read skill before deletion: {e}")
 
-            # Resolve UUID or name to UUID and delete the skill object folder
-            skill_uuid = skill_handler.resolve_to_uuid_or_error(uuid_or_name)
-            result = skill_handler.delete_object(skill_uuid)
-
-            # Update cache after deletion
+            # Update cache BEFORE deletion (fixes parent chain while object still exists)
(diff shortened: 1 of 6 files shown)


src/skillberry_store/fast_api/skills_api.py
```diff
@@ -277,11 +277,7 @@ def delete_skill(uuid_or_name: str):
             except Exception as e:
                 logger.warning(f"Could not read skill before deletion: {e}")
 
-            # Resolve UUID or name to UUID and delete the skill object folder
-            skill_uuid = skill_handler.resolve_to_uuid_or_error(uuid_or_name)
-            result = skill_handler.delete_object(skill_uuid)
-
-            # Update cache after deletion
+            # Update cache BEFORE deletion (fixes parent chain while object still exists)
```
eranra, line 280: I think that this line: 

skill_uuid = skill_handler.resolve_to_uuid_or_error(uuid_or_name) is missing in the new version?

eranra: I do not see this in all other places ... so maybe this is not needed/supported

eranra (approved):
approve

src/skillberry_store/fast_api/skills_api.py
erezh16: Line 268 of new version

erezh16:
The race condition was actually deleting the object before updating the cache. Fixed across all APIs.

PR #169, opened 2026-06-08
Update store SDK with admin ops

erezh16:
After I moved the admin ops to the Python backend, I forgot to update the SDK. Fix.


aviweit (approved):
I approve. Thanks.

PR #177, opened 2026-06-09
Stability Fixes

erezh16:
- **Updated client SDK**
- **Excluded skillberry_common packages from tests to avoid conflicts**
- **Removed str from extra to avoid generation of complex OpenAPI types**
- **Fixed FastMCP shutdown (VMCP server) when the server is running as a test fixture**


src/skillberry_store/schemas/manifest_schema.py
```diff
@@ -67,7 +67,7 @@ class ManifestSchema(BaseModel):
 
     @field_validator("extra", mode="before")
     @classmethod
-    def parse_extra_json_string(cls, value):
+    def parse_extra(cls, value):
```
aviweit, line 70: @erezh16 , can you please explain why you renamed this method ?

pytest.ini
```diff
@@ -1,5 +1,6 @@
 # pytest.ini
 [pytest]
 pythonpath  = . src
+norecursedirs = skillberry-common
```
aviweit, line 4: Please explain this need. Thanks

erezh16: Currently, running `make test` in `skillberry-store` also invokes the `llm-client` tests in `skillberry-common`. The `llm-client` tests have a separate `pytest.ini` configuration, but when it is used as part of the bigger `make test`, it is ignored, and cause the tests to fail.

src/skillberry_store/schemas/manifest_schema.py
erezh16: This validator operates on any content targeted  for `extra` - not just a JSON string. The name is misleading.

PR #183, opened 2026-06-10
Per-endpoint authentication for skill imports (api_key / login_url / OAuth discovery / anonymous default)

Ronen-Levy:
## Summary

Adds a provider-agnostic authentication layer for outbound skill-import fetches, configured per endpoint in `config.yaml`. Previously imports fetched remote URLs (e.g. GitHub) unauthenticated, which fails for private sources and hits unauthenticated rate limits on recursive/batch imports.

The only assumption is that the remote accepts an `Authorization: Bearer <token>` API token — nothing is GitHub-specific.

## Behavior

Interactive imports default to **anonymous** (no token, no auth lookup) — public sources work with zero setup. A checkbox in the importer UI (checked by default) toggles this.

When **not** anonymous, auth for the matched endpoint (host-equality or URL-prefix; longest match wins) resolves in order:

1. `api_key` → `Authorization: Bearer <key>`
2. caller's `X-Endpoint-Token` header → Bearer (the forced-reauthentication retry path)
3. `login_url` → `401 {error: "login_required", login_url}`; user logs in, retries with `X-Endpoint-Token`
4. **OAuth discovery** → probe `<host>/.well-known/oauth-authorization-server` then `/.well-known/openid-configuration`, use the discovered `authorization_endpoint` as the login URL

No matching endpoint → `API_KEY` env-var fallback (or anonymous). Supplying an `X-Endpoint-Token` forces non-anonymous, so retries are never stranded by the checkbox.

## Changes

**New**
- `src/skillberry_store/tools/endpoint_auth.py` — config model, cached YAML loader, URL matcher, `discover_login_url()` (RFC 8414 / OIDC), `ReauthRequired`, and `resolve_auth_headers()` (single entry point).
- `config.yaml.example` — annotated template.
- `config.yaml` — default (github.com → `login_url`, since GitHub has no discovery metadata).

**Modified**
- `tools/anthropic/importer.py` — `_auth_headers()` delegates to the resolver; `url` / `override_token` / `anonymous` threaded through `fetch_from_github` and `import_from_anthropic_skill`.
- `fast_api/skills_api.py` — `import-anthropic` and `detect-anthropic-skills` take `anonymous` (Form) + `X-Endpoint-Token` (Header); `ReauthRequired` → structured `401` via `_auth_exception_to_http()`.
- `fast_api/server.py` — load config at startup (logs path + entry count).
- `ui/src/components/AnthropicSkillImporter.tsx` — anonymous-login checkbox (default on); on `401 login_required` (single **and** batch) auto-opens the login URL and offers a token-paste retry that resends with `X-Endpoint-Token`. (Also fixes a pre-existing TS type error in `detectChildSkills`.)

## Backward compatibility

- No `config.yaml` / no match → unchanged (anonymous, or `API_KEY` env if set).
- Malformed/missing config → logged warning + treated as empty; never blocks startup.
- `pyyaml` already a dependency; no new deps.

## Testing

- Unit: all resolution paths (api_key, override token, login_url, discovery-found, discovery-none, API_KEY fallback, anonymous bypass), discovery mocked.
- Existing import/skills unit suite passes (35 tests).
- End-to-end: default (anonymous) detect succeeds with no token; `anonymous=false` → `401 login_required`; `X-Endpoint-Token` is forwarded to the provider.

## Out of scope

- Full OAuth handshake / device flow (only **discovery** of the login URL is included).
- Persisting/caching user-supplied `X-Endpoint-Token` values.
- Applying endpoint auth to non-import fetches (vmcp/vnfs call the service itself on localhost).


src/skillberry_store/tools/endpoint_auth.py
```diff
@@ -0,0 +1,285 @@
...
+import requests
+import yaml
+
+logger = logging.getLogger(__name__)
+
+# Auth modes
+MODE_API_KEY = "api_key"
+MODE_FORCED_REAUTH = "forced-reauthentication"
+
+_DEFAULT_CONFIG_FILENAME = "config.yaml"
```
eranra: On my machine the github credentials are saved in the default location ~/.config/gh/hosts.yml. 

@Ronen-Levy Can you add an automatic fallback to check this location andtry to use the keys from that file. This way the process will be 100% automatic for most of the users and most of the cases.

config.yaml
```diff
@@ -0,0 +1,16 @@
...
+# See config.yaml.example for the full annotated reference.
+#
+# Loaded once at startup (default ./config.yaml, or set SBS_CONFIG_FILE).
+# Per matched endpoint, auth resolves in order:
+#   1. api_key   -> sent as Authorization: Bearer <key>
+#   2. login_url -> forced-reauthentication: 401 + this URL; user retries with
+#                   header X-Endpoint-Token: <token>
+#   3. (neither) -> OAuth discovery (/.well-known/oauth-authorization-server)
+# No match -> API_KEY env var fallback (or anonymous).
+endpoints:
```
eranra: can we change this to be similar to this format:

$ cat ~/.config/gh/hosts.yml
github.ibm.com:
    user: ERANRA
    oauth_token: gho_rl4S...
    git_protocol: ssh
github.com:
    user: eranra
    oauth_token: gho_zJJ3R...
    git_protocol: ssh

config.yaml
```diff
@@ -0,0 +1,16 @@
+# Skillberry-store runtime configuration.
```
eranra: We need to add explanation on how to use that configuration somewhere in the docs ...

eranra: why do we need config.yaml and config.yaml.example .... we need to choose, either we provide config.yaml with the example, but remarked or we provide a config.yaml.example and do not provide the config.yaml and ask the users to copy+rename.

config.yaml.example
```diff
@@ -0,0 +1,50 @@
+# Skillberry-store runtime configuration.
+#
```
eranra: I have to say that from reading the text ... there are too many words for something as simple as configuration of auth :-) ---> can we make this more simple consice text.

src/skillberry_store/tools/endpoint_auth.py
```diff
@@ -0,0 +1,285 @@
+# Copyright 2025 IBM Corp.
+# Licensed under the Apache License, Version 2.0
+
+"""Per-endpoint authentication for outbound skill-import fetches.
+
+Configuration lives in a YAML file (``config.yaml`` by default, overridable via
```
eranra: since this config.yaml is used just for this purpose and not for any other configuration, please consider to rename the file into import_auth_config.yaml. All the other configurations are done in SKillbeyy-store through environment variables and not throught config file.

src/skillberry_store/tools/endpoint_auth.py
```diff
@@ -0,0 +1,285 @@
...
+_discovery_cache: Dict[str, Optional[str]] = {}
+
+
+def discover_login_url(url: str) -> Optional[str]:
+    """Discover a provider's login (authorization) endpoint for ``url``'s host.
+
+    Tries the RFC 8414 and OIDC well-known metadata documents and returns the
+    ``authorization_endpoint`` if present. Result (including a None "no
+    discovery" answer) is cached per host. Network/parse failures => None;
+    providers without metadata (e.g. github.com) simply return None.
```
eranra: I am asking myself if this is not an overkill for the first version ... we want to support github.com and maybe github.ibm.com at this stage, all the process for finding different URL for autentication can be pushed to a point in time that we will get a request from real users. 

I am not sure that we need all of that complexisy in the code for the requirements we have right now.

eranra (changes requested):
Hi @Ronen-Levy I did a bit of "over review" but I think that most of the comments that I made will make sense and shoulf be helpful

**Thanks for submitting this PR** and I really look to merge this today , or atr most on Shnday. 

Good morning :-)

src/skillberry_store/tools/endpoint_auth.py
Ronen-Levy: Done in `790ff13`. `gh_cli_token()` reads the `oauth_token` for the host from `~/.config/gh/hosts.yml` and uses it automatically when no `oauth_token`/`login_url` is configured. Caveat: when `gh` stores the token in the OS keyring (its default on many setups) there is no token in the file — we log a warning and fall through to anonymous.

config.yaml
Ronen-Levy: Done in `790ff13`. The config is now a host-keyed map matching that exact shape:
```yaml
github.com:
  user: eranra
  oauth_token: gho_xxxx
  git_protocol: https
```
(`login_url` is an optional extra key for the forced-reauth flow.)

config.yaml
Ronen-Levy: Done in `790ff13` — added a "Skill-import authentication" section to `docs/config-env-vars.md` with the format, the env var, and the resolution order.

Ronen-Levy: Done in `790ff13` — removed both `config.yaml` and `config.yaml.example`. There is now a single `import_auth_config.yaml` (with the example hosts commented out), so there is nothing to copy/rename.

config.yaml.example
Ronen-Levy: Done in `790ff13` — trimmed `import_auth_config.yaml` to a short header + one commented example, and moved the longer explanation into the docs.

src/skillberry_store/tools/endpoint_auth.py
Ronen-Levy: Done in `790ff13` — renamed to `import_auth_config.yaml`, env override `SBS_IMPORT_AUTH_CONFIG`. Agreed it is import-auth only; everything else stays on env vars.

src/skillberry_store/tools/endpoint_auth.py
Ronen-Levy: Agreed — removed the OAuth `.well-known` discovery entirely in `790ff13`. Resolution is now just `oauth_token` -> `login_url` -> `gh` CLI token -> anonymous, which covers github.com / github.ibm.com. We can revisit discovery if a real user asks for it.

Ronen-Levy:
Thanks for the thorough review @eranra — all points addressed in `790ff13`:

| Comment | Change |
|---|---|
| gh hosts.yml auto-fallback | Reads `oauth_token` from `~/.config/gh/hosts.yml`; warns + anonymous when token is in the keyring |
| gh-style config format | Config is now a host-keyed map: `user` / `oauth_token` / `git_protocol` (+ optional `login_url`) |
| Docs | Added a "Skill-import authentication" section to `docs/config-env-vars.md` |
| config.yaml vs .example | Removed both; single `import_auth_config.yaml` with commented examples — nothing to copy/rename |
| Too verbose | Trimmed the file to a short header + one example |
| Rename file | `import_auth_config.yaml`, env override `SBS_IMPORT_AUTH_CONFIG` |
| OAuth discovery overkill | Removed entirely; resolution is `oauth_token` → `login_url` → `gh` token → anonymous |

Ready for another look.

eranra:
@Ronen-Levy thanks for the fixes. All make sense and ready, can you please (a) rebase on main (b) lint  to pass the "[pull_request / pull-request (pull_request)](https://github.com/skillberry-ai/skillberry-store/actions/runs/27344925913/job/80801968343?pr=183)" and + (c)  fix the "security vulnerability"  to pass the [Code scanning results / CodeQL](https://github.com/skillberry-ai/skillberry-store/pull/183/checks?check_run_id=80802160771)

eranra:
Works very nicely. I tested both anonymous and authenticated modes. The anonymous execution hit the rate limit, while the authenticated execution completed successfully, as shown in the attached screenshot.

image

Ronen-Levy:
@eranra pushed `839bfe7` addressing all three of your asks:

- **(a) Rebased** on latest `main` (clean, no conflicts).
- **(b) Lint** — ran `black` over the project dirs; `black --check` passes locally.
- **(c) CodeQL** — fixed the high-severity *"uncontrolled data used in path expression"* in `detect_anthropic_skills` (`skills_api.py`). `folder_path` is now normalized with `os.path.realpath` before any filesystem access, only plain basenames are joined, and resolved entries are confined to the sanitized root. Verified legitimate skill folders are still detected.

The CI workflows for this commit are showing **`action_required`** (GitHub gates workflow runs on fork PRs until a maintainer approves them) — could you click **"Approve and run"** on the `pull_request` and `CodeQL` runs so the checks can execute? Thanks!

eranra (approved):
All greed. LGTM

PR #200, opened 2026-06-15
fix(skill-optimizer): correctly import YAML folded-block description from SKILL.md

### LINKED ISSUE #199: fix(skill-optimizer): optimized skill description shows '>' instead of full text
## Problem

When the Skill Optimizer plugin optimizes a skill, the resulting skill's **description field shows only \`>\`** in the UI, while the correct description is visible in the Additional Information (metadata) field.

## Root cause

The importer's \`parse_skill_metadata()\` used a single-line regex to capture the \`description\` field from SKILL.md frontmatter:

\`\`\`python
desc_match = re.match(r"^description:\s*(.+)$", line_stripped, re.IGNORECASE)
\`\`\`

When the optimizer agent wrote a YAML folded block scalar (\`description: >\`) — a format explicitly shown as a "Good example" in the knowledge file — the regex captured only \`>\`, leaving the description empty in the UI.

## Fix

- **importer.py**: Replace the manual regex parser with \`yaml.safe_load()\` applied to the extracted frontmatter block. This correctly handles \`>\`, \`|\`, quoted, and single-line YAML values.
- **plugin.py**: Add a fallback — if \`skill_description\` from the importer is empty, use \`opt_metadata["skill_description"]\` from \`required_outputs.json\` (which always contains the correct value).

## Knowledge improvements

- **01-skillberry-store-format.md**: Clarify that both single-line and folded block scalar descriptions are supported.
- **02-skill-best-practices.md**: Add "Explain WHY, Not Just WHAT" writing style section from the official Anthropic skill-creator guidance.

eranra:
Closes #199.

## Summary

- **Bug**: Optimized skill description showed `>` in UI instead of the full description text
- **Root cause**: `parse_skill_metadata()` regex only captured the first token on the `description:` line; YAML folded block scalars (`description: >`) were not handled
- **Fix**: Replace regex parsing with `yaml.safe_load()` on the extracted frontmatter block; add fallback to `opt_metadata["skill_description"]` from `required_outputs.json` as defense-in-depth
- **Knowledge**: Clarify both single-line and `>` formats are supported; add "Explain WHY not MUST" writing style guidance from official Anthropic skill-creator

## Changes

| File | Change |
|------|--------|
| `src/skillberry_store/tools/anthropic/importer.py` | Replace regex frontmatter parser with `yaml.safe_load()` |
| `plugins/.../plugin.py` | Fallback description from `opt_metadata["skill_description"]` when importer returns empty |
| `knowledge/01-skillberry-store-format.md` | Document both single-line and folded block scalar description formats |
| `knowledge/02-skill-best-practices.md` | Add "Explain WHY, Not Just WHAT" section from official skill-creator guidance |

Signed-off-by: Eran Raichstein <eranra@il.ibm.com>

### DIFF
--- a/node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/results.json
+++ b/node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/results.json
@@ -1,0 +1,1 @@
+{"version":"4.1.8","results":[[":src/skillberry_store/ui/src/hooks/useChangesMonitor.test.ts",{"duration":23.78659300000004,"failed":true}]]}
--- a/node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/results.json
+++ b/node_modules/.vite/vitest/da39a3ee5e6b4b0d3255bfef95601890afd80709/results.json
@@ -1,1 +1,0 @@
-{"version":"4.1.8","results":[[":src/skillberry_store/ui/src/hooks/useChangesMonitor.test.ts",{"duration":23.78659300000004,"failed":true}]]}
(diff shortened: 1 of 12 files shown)


eranra:
**Follow-up commit** pushed after repro with new optimization run.

**Root cause of "Anthropic skill imported from folder"**: The previous fix had the priority backwards — `skill_description or opt_metadata.get(...)` meant the generic placeholder string "Anthropic skill imported from folder" blocked the fallback entirely (it's truthy). The actual YAML failure was `yaml.safe_load` choking on a colon-containing plain scalar in the description (e.g. `description: Use this skill for operations: booking flights...`), silently returning `None` via the `except` block (which previously only `print()`-ed to stdout).

**Fix** (2367db2): Reverse the priority to `opt_metadata["skill_description"] or skill_description or ""` — the optimizer agent's `required_outputs.json` is always the authoritative source. Also replaced the silent `print()` with `logger.warning` so future YAML failures appear in server logs.

PR #208, opened 2026-06-18
Derive MCP tool content from stored manifest instead of live client

### LINKED ISSUE #207: MCP tool content endpoint spawns a live client connection to retrieve stored metadata
When fetching the content of an MCP-packaged tool, the endpoint currently establishes a live connection to the MCP server to read back tool metadata that was already captured in the manifest at registration time. This makes the endpoint brittle (it fails when the server is offline), slower than necessary, and adds an unused dependency on the `get_mcp_tools` / `mcp_content` helpers that only exist for this path.

bnayahu:
## Summary
- Removes the `get_mcp_tools` and `mcp_content` call path from the tool content endpoint
- Uses `mcp_content_from_manifest` (already available) to serve the response from stored data
- Eliminates an unnecessary 404 branch that could never be reached once manifest data is present

Closes #207

## Test plan
- [ ] Existing tools API tests cover the affected endpoint
- [ ] Run `pytest src/skillberry_store/tests/`

### DIFF
--- a/src/skillberry_store/fast_api/tools_api.py
+++ b/src/skillberry_store/fast_api/tools_api.py
@@ -49,9 +49,7 @@ from skillberry_store.utils.utils import (
 )
 from skillberry_store.utils.python_utils import extract_docstring
 from skillberry_store.fast_api.server_utils import (
-    get_mcp_tools,
     mcp_json_converter,
-    mcp_content,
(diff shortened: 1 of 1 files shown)


bnayahu:
CI is failing due to connectivity issues to HF.

eranra (approved):

PR #211, opened 2026-06-18
Filter vMCP server list by associated skill

### LINKED ISSUE #210: GET /vmcp_servers/ has no server-side filtering by associated skill
The vMCP list endpoint returns all registered virtual MCP servers with no way to narrow results to those associated with a specific skill. Callers that need only the servers for a given skill must fetch the full list and filter client-side, which is inefficient and couples clients to the internal data shape.

bnayahu:
## Summary
- Adds an optional `skill_uuid` query parameter to `GET /vmcp_servers/`
- When provided, the response includes only servers whose `skill_uuid` field matches
- When omitted, behaviour is unchanged (all servers returned)

Closes #210

## Test plan
- [ ] `test_vmcp_api_filter.py` covers filtered and unfiltered responses
- [ ] Run `pytest src/skillberry_store/tests/test_vmcp_api_filter.py`

### DIFF
--- a/src/skillberry_store/fast_api/vmcp_api.py
+++ b/src/skillberry_store/fast_api/vmcp_api.py
@@ -125,7 +125,8 @@ def register_vmcp_api(
                 servers = result["virtual_mcp_servers"]
                 result = {
                     "virtual_mcp_servers": {
-                        k: v for k, v in servers.items()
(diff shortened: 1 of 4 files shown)


eranra (approved):
LGTM

PR #214, opened 2026-06-18
Support dynamically-populated and cascading dropdowns in plugin action forms

### LINKED ISSUE #213: PluginActionForm only supports static fields — no dynamic or cascading dropdowns
Plugin action forms render fields from a JSON Schema but have no mechanism for a plugin to declare that a dropdown's options should be fetched from an API endpoint at runtime. There is also no way to express that one field's options depend on the selected value of another field (cascading selects). Plugins that need to present context-aware option lists must either hard-code values into the schema or build their own form outside the standard plugin UI contract.

bnayahu:
## Summary
- Reads `x-options-from` on a JSON Schema property to fetch dropdown options from a URL at form-open time
- Supports `x-option-label` / `x-option-value` to map arbitrary response fields to label/value pairs
- Supports `x-exclude-tags` to filter items that carry specific tags from the options list
- Supports `x-depends-on` to make a dropdown's options reload whenever a named parent field changes; the dependent field is cleared and disabled until the parent has a value

Closes #213

## Test plan
- [ ] Open a plugin action form whose schema includes `x-options-from` and verify options load
- [ ] Verify a dependent dropdown reloads when its parent value changes
- [ ] Verify the dependent dropdown is disabled when the parent has no value

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/skillberry_store/ui/src/components/PluginActionForm.tsx
+++ b/src/skillberry_store/ui/src/components/PluginActionForm.tsx
@@ -1,7 +1,7 @@
 // Copyright 2025 IBM Corp.
 // Licensed under the Apache License, Version 2.0
 
-import { useState } from 'react';
+import { useState, useEffect } from 'react';
 import {
   Modal,
   ModalVariant,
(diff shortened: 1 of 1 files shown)


eranra (approved):
LGTM

PR #222, opened 2026-06-18
fix: make arg_convert produce JSON-native MCP call arguments

### LINKED ISSUE #221: MCP tool execution corrupts string arguments (arg_convert quotes JSON args)
## Summary

Executing an MCP-packaged tool through the store sends **corrupted arguments** to the backend MCP server. A string parameter such as `user_id="sara_doe_496"` is delivered to the server as `'"sara_doe_496"'` — with literal embedded quote characters — so lookups fail with errors like *"user not found"*.

The same tool call issued **directly** against the backend (e.g. via the simulation-harness test client) works correctly, because that path sends raw JSON and never goes through the store's argument conversion.

## Root cause

`arg_convert` in `src/skillberry_store/modules/file_executor.py` was written to emit **Python source-code literals** (a string becomes a quoted `"value"`). Its only caller, however, is the MCP execution path (`execute_python_file_in_mcp_server`), which uses the return value to build the `arguments` dict for `session.call_tool(...)`. There the value must be a **JSON-native object**, not a source literal — so the quoting is a bug.

Additionally, the store stores parameter types using JSON-schema names (`"string"`, `"integer"`, …), which never matched `arg_convert`'s `"str"`/`"int"` branches, so values fell through to a guess-coercion fallback.

## Impact

The fallback mishandles several JSON types when building MCP call arguments:

| Input | Declared type | Sent to MCP server | Expected |
|-------|---------------|--------------------|----------|
| `"sara_doe_496"` | string | `'"sara_doe_496"'` | `"sara_doe_496"` |
| `"496"` | string | `496` (int) | `"496"` |
| `"12:30"` | string | `datetime.time` (not JSON-serializable) | `"12:30"` |
| `"true"` | boolean | `'"true"'` (quoted string) | `true` |

## Why it wasn't caught earlier

- `arg_convert` has exactly one caller and had no unit tests.
- Existing regular (non-simulated) MCP tools are created without a `params` field, so they bypass `arg_convert` entirely (the no-parameter shortcut in `execute_python_file_in_mcp_server`).
- Simulated tools (Simulate plugin) are the first to carry typed params through this path, which is what surfaced the bug.

## Fix

Make `arg_convert` type-driven: map JSON-schema/Python type names to JSON-native values (str/int/float/bool, containers passed through) with no source-quoting and no guess-coercion. Add unit coverage.

bnayahu:
Fixes #221

## Problem

Executing an MCP-packaged tool through the store corrupts string arguments. A parameter like `user_id="sara_doe_496"` reaches the backend MCP server as `'"sara_doe_496"'` (literal embedded quotes), so lookups fail with *"user not found"*. The same call issued directly against the backend works, because that path sends raw JSON and never goes through the store's argument conversion.

## Root cause

`arg_convert` (`src/skillberry_store/modules/file_executor.py`) was written to emit **Python source-code literals** (a string becomes a quoted `"value"`). Its only caller is the MCP path (`execute_python_file_in_mcp_server`), which uses the result to build the `arguments` dict for `session.call_tool(...)` — where the value must be a **JSON-native object**, not a source literal. The store also stores parameter types as JSON-schema names (`"string"`, `"integer"`, …), which never matched `arg_convert`'s `"str"`/`"int"` branches, so values fell through to a guess-coercion fallback.

This also produced three latent bugs: numeric-looking string IDs (`"496"`) coerced to `int`, time-looking strings (`"12:30"`) turned into non-JSON-serializable `datetime.time`, and booleans quoted as strings.

## Why it wasn't caught earlier

- `arg_convert` has exactly one caller and had no unit tests.
- Regular (non-simulated) MCP tools are created without a `params` field, so they bypass `arg_convert` entirely via the no-parameter shortcut.
- Simulated tools (Simulate plugin) are the first to carry typed params through this path, which surfaced the bug.

This means the change **cannot regress existing regular-MCP behavior** — no existing regular-MCP tool reaches this function.

## Fix

Rewrite `arg_convert` to be type-driven: map JSON-schema/Python type names to JSON-native values (str/int/float/bool, containers passed through) with no source-quoting and no guess-coercion.

## Testing

- New parametrized unit test `test_arg_convert_produces_json_native_values` (14 cases) — failed 9/14 before, all pass after.
- `test_file_executor.py` under local execution: 42 pass, 1 fail (`test_execute_modes[docker_mode]`, requires Docker — unrelated to this change).
- Dependency / tau2 / env_id suites: 33 pass. Simulate plugin manifest/synth: 8 pass.

Note: the live end-to-end simulation path (Docker + harness + LLM key) was not exercised; the fix is verified at the unit level and by tracing the exact `call_tool` argument construction.

### DIFF
--- a/src/skillberry_store/modules/file_executor.py
+++ b/src/skillberry_store/modules/file_executor.py
@@ -43,49 +43,38 @@ def get_distribution(module_name: str) -> str:
 
 
 def arg_convert(arg_name, arg_type):
-    arg_str = str(arg_name)  # Ensure the argument is treated as a string
-    if arg_type is not None:
-        if arg_type == "str":
-            return f'"{arg_str}"'
-        elif arg_type == "int":
-            try:
-                int_val = int(arg_str)
-                # Check if the string representation matches to avoid float->int conversion
-                if str(int_val) == arg_str:
-                    return int_val
-            except Exception as e:
-                raise ValueError(f"Cannot convert '{arg_str}' to int {e}")
+    """Coerce a parameter value to the JSON-native type declared in the tool schema.
 
-    try:
-        int_val = int(arg_str)
-        # Check if the string representation matches to avoid float->int conversion
-        if str(int_val) == arg_str:
-            return int_val
-    except ValueError:
-        pass
-
-    # Try to convert to float
-    try:
-        float_val = float(arg_str)
-        return float_val
-    except ValueError:
-        pass
-
-    try:
-        return float(arg_str)
-    except ValueError:
-        pass
-
-    try:
-        parts = arg_str.split(":")
-        if len(parts) == 2:
-            return datetime.datetime.strptime(arg_str, "%H:%M").time()
-        elif len(parts) == 3:
-            return datetime.datetime.strptime(arg_str, "%H:%M:%S").time()
-    except ValueError:
-        pass
-
-    return f'"{arg_str}"'
+    This is used to build the ``arguments`` dict for an MCP ``call_tool`` request,
+    so the returned value must be a JSON-serializable Python object
+    (str/int/float/bool/list/dict) — NOT a source-code literal. The accepted type
+    names cover both the JSON-schema vocabulary the store stores and emits in the
+    MCP stub ("string", "integer", "number", "boolean", "array", "object") and the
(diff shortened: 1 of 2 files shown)


eranra (approved):
LGTM

PR #223, opened 2026-06-18
feat: add Documentation Generator plugin

### LINKED ISSUE #201: Documentation generator & enricher for skills, tools, and snippets
## Problem

The store curates skills, tools, and snippets and has a rich set of plugins to *assess* them (security, SAST, provenance, dedupe, evaluation) and *improve their code* (optimizer). But nothing maintains their **documentation** — the human-readable description, usage, parameters, and examples that determine whether anyone can actually understand and reuse an object.

Today documentation depends entirely on whatever the author or import source happened to write, and it drifts immediately:

- **Imported objects arrive with thin or missing docs.** A skill imported from a GitHub URL keeps only whatever `SKILL.md` / docstrings existed upstream; a tool imported from an MCP server may have just a one-line description; snippets often have none. Quality is inconsistent and outside the store's control.
- **Docs and code diverge over time.** When the optimizer rewrites a tool's code, or a tool's parameters change, the description and examples are not updated — so the documentation silently becomes wrong, which is worse than missing.
- **No consistent shape across object types.** Skills, tools, and snippets each end up documented (or not) in ad-hoc ways, making them hard to scan, search, compare, or present uniformly in the UI.
- **The information needed to write good docs already exists but isn't synthesized.** The code, parameter schemas, tool/snippet content, tags, and even other plugins' findings are all in the store — yet nothing turns them into a clear, complete description, usage notes, and examples.

The result is an inconsistent, decaying documentation layer that undercuts the store's core purpose: making curated skills/tools/snippets discoverable and reusable.

## Who is affected

- **A user browsing or searching the store** — relies on descriptions/examples to decide whether an object fits their need; thin or stale docs make objects effectively unusable even when the code is good.
- **A user importing or authoring objects** — wants the store to bring their object up to a consistent documentation standard without hand-writing boilerplate.
- **Downstream consumers (agents, MCP clients)** — consume tool/skill descriptions and parameter docs directly; poor docs degrade tool selection and invocation.
- **The store operator** — wants a uniform, trustworthy documentation standard across the whole catalog.

## Desired outcome

A way to **generate documentation where it's missing and enrich/refresh it where it's thin or stale**, for all three object types, producing a consistent documentation shape across the catalog. At minimum it should be able to:

1. **Generate** a clear description, usage/when-to-use guidance, parameter/input documentation, and at least one example for an object that lacks them — derived from the object's own code, content, parameter schemas, and metadata already in the store.
2. **Enrich** existing-but-thin documentation (e.g. expand a one-line tool description, add missing parameter docs or examples) without discarding good author-written content.
3. **Refresh / detect drift** when an object's code or interface changes (e.g. after the optimizer rewrites it, or a tool's parameters change), flagging or regenerating documentation that no longer matches the code.
4. **Apply consistently across skills, tools, and snippets**, accounting for each type's structure (a skill's referenced tools/snippets, a tool's module + parameters, a snippet's inline content).
5. **Be reviewable and non-destructive** — surface proposed documentation for the user before overwriting, and record what was generated/changed so author intent isn't silently lost.

## Out of scope (for this issue)

- The mechanism (plugin vs. core, which model, exact storage fields, UI).
- Whether generation runs automatically on import vs. on demand vs. both.
- Translating/localizing documentation into other languages.

## Open questions

- Should generated docs **replace**, **augment**, or live **alongside** author-written docs (e.g. a clearly-marked "generated" section)?
- Should enrichment run automatically on import/optimize events, only on demand, or both — and should it ever overwrite without review?
- How is documentation **drift** detected — compare against a hash of the code/interface captured when docs were last written (similar to the provenance baseline approach)?
- What is the canonical documentation shape per object type, and does the existing schema (`description`, `extra`, parameter definitions) hold it, or is a dedicated field needed?

Ronen-Levy:
## Summary

Adds a new `skillberry-plugin-doc-generator` plugin that **generates documentation where it's missing, enriches thin docs without discarding author content, and detects drift** when an object's code or interface changes — for **skills, tools, and snippets**, in one consistent shape.

The store already has plugins to *assess* objects (security, SAST, provenance, dedupe, evaluation) and to *improve code* (optimizer), but nothing maintains their **documentation** — the description, usage, parameters, and examples that decide whether anyone can understand and reuse an object. This plugin fills that gap.

Resolves #201.

## What it does

Three operations, one code path:
- **Generate** — produce docs for an object that lacks them, derived from its own code, content, parameter schema, tags, and (for skills) referenced children.
- **Enrich** — expand thin author docs **without discarding** them (`mode: enriched`); sufficient author content is kept verbatim (`mode: kept`).
- **Refresh / drift** — compare an object's current `source_fingerprint` (code/params/references) against the applied docs, flag drift, and propose regenerated docs.

Every object type gets the **same documentation shape**, stored under `extra["documentation"]`:

| Field | Meaning |
|-------|---------|
| `description` | What the object is / does |
| `when_to_use` | Usage guidance |
| `parameters` | Per-input docs (`name`, `type`, `required`, `description`) |
| `examples` | At least one usage example |
| `mode` | `generated` / `enriched` / `kept` |
| `notes` | Honesty notes about what couldn't be inferred |

## Design

Follows the patterns established by the provenance and SAST plugins:

- **Non-destructive & review-before-apply.** Results are stored under `extra["documentation"].proposed` until applied (`apply=true`), which promotes them to `.current`. The raw object `description` field is never overwritten. On import, the plugin auto-*proposes* docs for objects that lack them (best-effort, never auto-applies, never blocks an import).
- **LLM wired exactly like the security evaluator plugin (`llm-switchboard`).** Documentation is produced by a frontier model: `get_llm(LLM_PROVIDER)` → `Client(model_name=LLM_MODEL)` → `await client.generate_async(prompt=...)`. No specific model is hardcoded; provider/model come from `LLM_PROVIDER` (default `openai.async`) / `LLM_MODEL` (default `gpt-4`). `llm-switchboard[litellm]` is a hard dependency. If the LLM cannot be initialized the plugin is **disabled** (`is_enabled()` is `False`) and its endpoints return `503`, exactly like the security plugin. The `mode` label is decided from the author's existing description so the non-destructive guarantee holds (sufficient author content is kept verbatim).

## API

Mounted at `/plugins/doc_generator` (and `/api/plugins/doc_generator/...` for the UI):

- `POST /generate` — `{ "object_type": "skill|tool|snippet", "uuid": str, "apply"?: bool, "only_if_missing"?: bool }`
- `POST /refresh`  — `{ "object_type": "skill|tool|snippet", "uuid": str }`

Both return `{ success, message, data }`. Invalid/missing input returns a clean **400** with a helpful message (rather than a raw pydantic 422), so the store's generic action form behaves well even when its enum field isn't explicitly changed. When the LLM is unavailable, both endpoints return **503**.

## Files

- **New:** `plugins/skillberry-plugin-doc-generator/` — `pyproject.toml`, `README.md`, `plugin.py`, `models.py`, and tests.
- **Changed:** root `pyproject.toml` — registers the plugin in the `plugins-all` group, adds a `plugin-doc-generator` extra, and the editable `[tool.uv.sources]` path (required for CI to install it).

## Testing

- `pytest plugins/skillberry-plugin-doc-generator/tests -q` → **30 passed** (prompt building, JSON parsing + bad/empty-response handling, mode classification, enablement/503 gating, 400-validation, plugin persistence, drift, event handlers). The `llm-switchboard` client is mocked the same way the security plugin's tests mock it (an `AsyncMock` returns canned JSON).
- Full-repo collection: **814 tests, 0 errors**.
- `black --check` clean.

## Notes / out of scope

- Tests dir intentionally has **no `__init__.py`** (avoids the cross-plugin `tests` package collision fixed in #198).
- No UI changes — the plugin declares its actions via `get_ui_config()`, rendered by the store's generic action form.


eranra (approved):
LGTM

PR #229, opened 2026-06-21
fix(plugins): read credentials from settings.json env block and forward it fully

### LINKED ISSUE #228: Plugins ignore the ~/.claude/settings.json env block, so configured credentials read as "Missing credentials"
The `skill-optimizer` and `anthropic-skill-generator` plugins load `~/.claude/settings.json` to discover Anthropic credentials, but they look for **non-standard top-level keys** (`apiKey`, `baseUrl`, `authToken`, `model`):

```python
self._claude_settings.get("baseUrl")
self._claude_settings.get("authToken")
```

The real Claude Code `settings.json` schema nests these under an **`env`** block using the standard variable names:

```json
{
  "env": {
    "ANTHROPIC_BASE_URL": "https://...",
    "ANTHROPIC_AUTH_TOKEN": "sk-...",
    "ANTHROPIC_MODEL": "claude-opus-4-8"
  }
}
```

Because the lookup keys don't match, `_check_credentials()` returns `False` and the plugin shows **"Missing credentials"** even when `settings.json` is correctly configured.

A second problem: even when credentials are found, `_build_claude_env()` forwards only a hardcoded 4 variables, so model selection and `CLAUDE_*` flags (`ANTHROPIC_SMALL_FAST_MODEL`, `CLAUDE_CODE_SUBAGENT_MODEL`, …) are dropped before the agent runs.

**Expected:** credentials resolve from the standard `env` block, and the whole `env` block is forwarded to the Claude Code agent.

PrielHaz:
Closes #228

The `skill-optimizer` and `anthropic-skill-generator` plugins read `~/.claude/settings.json` but looked for non-standard top-level keys (`apiKey`/`baseUrl`/`authToken`). Real Claude Code settings nest these under an `env` block using the standard `ANTHROPIC_*` names, so credentials were never found and the plugins reported "Missing credentials" despite being configured.

## Changes
- Resolve credentials from the `settings.json` `env` block using the standard `ANTHROPIC_API_KEY` / `ANTHROPIC_BASE_URL` / `ANTHROPIC_AUTH_TOKEN` names.
- Forward the **entire** `env` block to the Claude Code agent instead of a hardcoded subset, so models and `CLAUDE_*` flags propagate (`ANTHROPIC_MODEL`, `ANTHROPIC_SMALL_FAST_MODEL`, `CLAUDE_CODE_SUBAGENT_MODEL`, …).
- Process `ANTHROPIC_*` / `CLAUDE_*` environment variables still override the settings file; per-run `agent_env` still wins over both.
- Clarify the `agent_env` UI field: `PluginActionForm` now honors a field `title` for the default input (previously it showed the raw property name), and both plugins relabel the field to "Environment overrides (optional)" with a description noting that `settings.json` + server env are already loaded automatically.
- Isolate the credential tests from the dev machine's real `settings.json`, and add coverage for env-block parsing and full forwarding.

## Testing
- `pytest` for both plugins: **49 passed**.

## Files
- `plugins/skillberry-plugin-skill-optimizer/.../plugin.py`
- `plugins/skillberry-plugin-anthropic-skill-generator/.../plugin.py`
- `plugins/skillberry-plugin-skill-optimizer/tests/test_skill_optimizer_plugin.py`
- `src/skillberry_store/ui/src/components/PluginActionForm.tsx`

### DIFF
--- a/plugins/skillberry-plugin-anthropic-skill-generator/src/skillberry_plugin_anthropic_skill_generator/plugin.py
+++ b/plugins/skillberry-plugin-anthropic-skill-generator/src/skillberry_plugin_anthropic_skill_generator/plugin.py
@@ -150,75 +150,61 @@ class SkillberryPluginAnthropicSkillGenerator(PluginBase):
         else:
             logger.debug(f"Claude settings file not found at {settings_path}")
     
-    def _check_credentials(self) -> bool:
-        """Check if Anthropic credentials are configured."""
-        # Check for direct API key in environment
-        if os.getenv("ANTHROPIC_API_KEY"):
+    @staticmethod
+    def _has_api_access(source: Any) -> bool:
+        """True if `source` (a mapping) carries Anthropic API credentials."""
+        if source.get("ANTHROPIC_API_KEY"):
             return True
-        
-        # Check for proxy/gateway configuration in environment
-        if os.getenv("ANTHROPIC_BASE_URL") and os.getenv("ANTHROPIC_AUTH_TOKEN"):
+        if source.get("ANTHROPIC_BASE_URL") and source.get("ANTHROPIC_AUTH_TOKEN"):
             return True
-        
-        # Check for credentials in Claude settings file
-        if self._claude_settings:
-            # Check for API key in settings
-            if self._claude_settings.get("apiKey"):
-                return True
-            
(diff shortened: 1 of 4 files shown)


eranra (approved):
LGTM

PR #239, opened 2026-06-21
feat: add DAST Scanner plugin (dynamic, detect-and-report)

### LINKED ISSUE #238: DAST Scanner: dynamically test a skill's runtime behavior, not just its static code
## Problem

The store's security analysis today is entirely **static** — SAST, provenance, and dependency tracking all *read* an object without ever running it. None of them can answer the dynamic question that matters most for trust: **when this skill actually executes with hostile or unexpected input, what does it do?** — does it reach out to the network, spawn a subprocess, touch the filesystem, leak secrets, or crash?

Static analysis fundamentally can't answer that. Behavior depends on inputs and runtime paths that only surface when the code runs.

This gap matters because a skill is reusable **code**, not just an agent-mediated behavior. Any external component — an agent, a developer, another tool — can load a skill and invoke its functions directly. So the runtime surface that needs testing is **every externally-invocable entry point** of a skill (its registered tools *and* the other callables its modules expose), not only the path an agent happens to take.

## Who is affected

- **A user importing or reusing a skill** — wants assurance about what the skill *does at runtime*, beyond what static scanners can see.
- **The store operator** — wants a dynamic, behavior-based security signal alongside the existing static ones, to catch issues (unexpected egress, subprocess execution, data exfiltration shapes) that static analysis misses.
- **Downstream consumers (agents, MCP clients)** — invoke a skill's tools at runtime; behavior under adversarial input is exactly their risk surface.

## Desired outcome

A way to **dynamically test a skill** — exercise its entry points with adversarial/unexpected inputs and observe the resulting runtime behavior — producing a security/behavior report that complements the static analyzers. At minimum it should:

1. Identify the skill's externally-invocable entry points.
2. Drive them with adversarial inputs.
3. Observe and report runtime effects (e.g. network access, subprocess execution, filesystem access, crashes, secret leakage).
4. Be safe and on-demand — run in a controlled/sandboxed manner, never as a blocking step on import.
5. Report honestly about coverage and confidence (what was and wasn't exercised).

## Out of scope (for this issue)

- The mechanism and engineering approach (how entry points are discovered, how inputs are generated, how execution is sandboxed, what tooling is used). These will be decided in the implementing pull request.

## Relationship to existing work

Complements — does not replace — the static analyzers (SAST, provenance, dependency-tracker): those inspect code at rest; this exercises it and observes behavior.

Ronen-Levy:
## What

Adds a **Dynamic Application Security Testing (DAST)** plugin. The store today has only *static* security analysis (SAST, provenance, dependency-tracker). This adds the *dynamic* counterpart: on demand it **discovers a skill's invocable entry points, exercises each with adversarial inputs, observes the runtime effects, and records findings + coverage** in a non-destructive `extra["dast"]` block.

Closes #238.

## How it works

- **Discovery (static).** Tier 1 = registered tools (driven by their param schema). Tier 2 = AST-discovered public functions / classes / `__main__` in the skill's modules. Tier 3 dynamic dispatch is acknowledged and reported as a coverage caveat (`discovery: "static (Tier-3 dynamic dispatch may be missed)"`).
- **Exercise.** The **MCP surface** is driven through a benign **vMCP twin**; the **direct surface** runs through the store's own `FileExecutor` with an observe-shim **prepended** to the code (no env/mount cooperation needed; events emitted as marker-prefixed JSON on stdout — the one channel that survives the Docker exec path).
- **Observe — detect-and-report only.** The shim is pass-through-and-log: it records network egress / subprocess / filesystem / secret-leak / crash effects but **never blocks** them. This ceiling is surfaced in the report (`observation: "detected, not prevented"`). Run only on operator-accepted networks. (True lockdown is a follow-up gated on a runspace capability request.)
- **Report.** `extra["dast"]` carries `coverage`, `entry_points`, `findings`, and a `summary` that separates **exercised** (calls executed) from **findings** (real issues), split by MCP vs direct surface. Plus `dast:` summary tags.

## Scope & bounding (env knobs)

- `DAST_SCOPE` — cumulative: `mcp` ⊆ `registered` ⊆ `discovered` (default `mcp`). `registered` adds direct Tier-1 calls; `discovered` adds Tier-2.
- `DAST_GENERATOR` — selects the pluggable input-generator engine.
- `DAST_MAX_CASES`, `DAST_MAX_ENTRY_POINTS`, `DAST_EXEC_TIMEOUT` — keep scans fast and bounded (per-case timeout via a bounded daemon thread, so a hanging tool can't stall a scan).
- On-demand only — registers **no** event handler.

## Optional engine (not bundled)

The input-generator engine is **optional and pluggable**, mirroring the SAST + bandit pattern. Core dependency is just `requests`. `hypothesis` is provided as an adapter but ships **only as an optional extra** (`pip install 'skillberry-plugin-dast[hypothesis]'`) — it is **not** a hard dependency and is **not** added to the store's deps. With no engine installed, the plugin reports **disabled** and `/scan` returns **503**.

## UI

**Backend-only PR.** Results are the `/scan` response summary plus the persisted `extra["dast"]` block and `dast:` tags — both render in the existing object view via the generic plugin action form. The `/scan-status` progress endpoint ships, but the UI polling for a live in-progress label is intentionally deferred to a follow-up (it would have required removing async-job types other plugins depend on).

## Verification

- `uvx black --check --target-version py311 plugins/skillberry-plugin-dast` — clean.
- `uv run pytest plugins/skillberry-plugin-dast/tests -q` — **56 passed** (offline; real Docker/vMCP gated behind `DAST_LIVE`).
- `uv run pytest --co -q` — full-repo collection error-free (#198 guard); `dast` entry point loads.
- Live: scanned a real skill with `DAST_SCOPE=registered` → exercised both MCP (96 calls) and direct registered (92 calls across 20/20 endpoints) surfaces, persisted `extra["dast"]`.

### DIFF
--- a/plugins/skillberry-plugin-dast/README.md
+++ b/plugins/skillberry-plugin-dast/README.md
@@ -1,0 +1,115 @@
+# skillberry-plugin-dast
+
+**DAST** (Dynamic Application Security Testing) — dynamically tests a **skill** by
+discovering its externally-invocable entry points, exercising each with
+adversarial inputs, and **observing the effects** (MCP calls, network egress,
+subprocess spawns, filesystem access). Results are recorded under
+`extra["dast"]` + `dast:` tags.
+
+This is the **dynamic** counterpart to the store's static analyzers (SAST,
+provenance, dependency-tracker): instead of reading code, it *runs* it with
+hostile inputs and watches what it actually does.
+
+## Threat model
+
+A skill is reusable **code**, not just its agent-mediated behavior — any external
+component (an agent, a developer, another tool) can load it and call its functions
+directly. So DAST targets **all enumerable entry points**, not just agent usage.
+
+## Entry points exercised (Phase 1)
+
+| Tier | What | Source |
+|------|------|--------|
+| **Tier 1** | registered tools (name + parameter schema) | tool metadata |
+| **Tier 2** | public top-level functions / classes / `__main__` | AST of the skill's modules |
+
+Tier 2 functions that shadow a registered tool are de-duped into Tier 1.
+
+## How it works
+
+1. **Discover** entry points (Tier 1 from tool metadata, Tier 2 via AST).
+2. **Generate inputs** for each via the optional Hypothesis engine — strategies
+   mapped from each parameter's declared type, plus missing-required cases
+   (deterministic; requires the engine, see below).
+3. **Exercise** each case via the store's `FileExecutor` (Docker exec sandbox),
+   with a pass-through-and-log **observe-shim** prepended to the code that
+   records `socket`/`requests`/`httpx`/`urllib`/`subprocess`/`os.system`/`open`.
+4. **Observe MCP** via a **benign vMCP twin** (`VirtualMcpServer`) the skill calls
+   instead of the real server — faithful execution + per-call logging.
+5. **Report** findings + coverage to `extra["dast"]`.
+
+## Honest ceilings (in every report's `coverage`)
+
+- **Discovery is static** → dynamically-dispatched entry points (Tier 3) may be
+  missed → `coverage.exercised = N/M`.
+- **Observation is detect-and-report, not prevent** → egress/effects *happen* and
+  are recorded, not blocked → `coverage.observation = "detected, not prevented"`.
+  **Run only against an operator-accepted network/sandbox.** True network lockdown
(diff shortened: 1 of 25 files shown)


eranra:
@Ronen-Levy  LGTM

eranra (approved):

PR #241, opened 2026-06-22
feat(tools): MCP-friendly tool creation from code string + curate MCP surface

### LINKED ISSUE #240: Tool creation impossible over MCP — endpoints require a file upload (422)
The store mounts every REST endpoint as an MCP tool via `fastapi-mcp` (`/control_sse`). An MCP tool call carries only JSON args — there is no channel for a `multipart`/octet-stream **file body** — so endpoints whose body is a required file upload receive a `null` body and fail with HTTP 422.

### Observed
A Runspace agent connected to the store MCP could create **snippets** and **skills** (string fields) but **could not create a tool**: `create_tool` and `add_tool_from_python` take the source as `module`/`tool: UploadFile = File(...)`, so every MCP call returned `422 body.tool: Field required (input: null)`.

### Fix
Add an MCP-friendly tool-creation endpoint that takes the source as a JSON string — `POST /tools/add_code` (see #241), reusing the docstring-extraction logic of `/tools/add`.

> Broader MCP-surface curation (dropping endpoints that can't be MCP tools at all) is tracked in a separate issue.

PrielHaz:
Closes #240

Tool creation over the control-plane MCP was impossible: `create_tool` and `add_tool_from_python` take the source as a required `UploadFile` (multipart), and an MCP tool call carries only JSON args — no file body — so the field arrives `null` and the call fails with HTTP 422. Agents could create snippets/skills (string fields) but never a real **tool** record.

### Changes
- **`POST /tools/add_code`** — takes the Python source as a JSON `code` string (plus optional `selected_func` / `update` / `module_name`) and reuses `add_tool_from_python`'s docstring extraction + registration. Works over MCP.
- **Curate the MCP surface** — pass `exclude_operations` to `FastApiMCP` to drop the endpoints that require a file upload and therefore can't be MCP tools: `create_tool`, `add_tool_from_python`, `restore_all_data`. The REST endpoints are unchanged; they're just no longer offered as (always-failing) MCP tools.

### Testing
In-process tests for `/tools/add_code`: extracts the manifest from the docstring and persists the string as the module file; honors `module_name`; requires `code` (422 otherwise). **3 passing.**

### DIFF
--- a/src/skillberry_store/fast_api/tools_api.py
+++ b/src/skillberry_store/fast_api/tools_api.py
@@ -6,6 +6,7 @@ import traceback
 from typing import Optional, Annotated, Dict, Any, List
 from fastapi import FastAPI, HTTPException, File, UploadFile, Query, Request
 from fastapi.responses import PlainTextResponse
+from pydantic import BaseModel
 
 from skillberry_store.modules.lifecycle import LifecycleState
 from skillberry_store.schemas.tool_schema import ToolSchema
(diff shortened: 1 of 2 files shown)


eranra (approved):
LGTM.

PR #243, opened 2026-06-22
feat(mcp): curated MCP surface (allow-list of operations)

### LINKED ISSUE #242: Expose a curated MCP surface (allow-list), not every REST endpoint
The control-plane MCP (`/control_sse`) auto-generates an MCP tool for **every** REST endpoint. Many of those don't belong on the agent surface — `health`, `readiness`, `metrics`, admin `backup`/`restore`/`purge`, `changes` polling, provenance `check`/`recheck` — and some **can't be MCP tools at all** (file-upload endpoints, which 422 because an MCP call has no file body; see #240).

### Proposal
Expose an explicit **allow-list** (`fastapi-mcp` `include_operations`) of just the content operations agents use. Starting list:

**Tools:** list, get, get_module, add_tool_from_code (#241), update, delete, execute, search
**Snippets:** create, list, get, update, delete, search
**Skills:** create, list, get, update, delete, search

Everything else (health/metrics/admin/changes/provenance/file-uploads/plugin-internal endpoints) is dropped from the MCP. Adding an `operation_id` to the list re-exposes it. (vmcp/vnfs can be added if agents need to manage them.)

PrielHaz:
Closes #242

The control-plane MCP auto-generated a tool for **every** REST endpoint, including ones that don't belong on the agent surface (health, readiness, metrics, admin backup/restore/purge, change polling, provenance, plugin-internal job endpoints) or can't be MCP tools at all (file-upload endpoints → 422, since an MCP call has no file body; see #240).

### Change
Switch `FastApiMCP` to an explicit **`include_operations` allow-list** — the content operations agents actually call over MCP:
- **Tools:** list, get, get_module, `add_tool_from_code`, update, delete, execute, search
- **Snippets:** create, list, get, update, delete, search
- **Skills:** create, list, get, update, delete, search
- **VMCP servers:** create, list, get, update, delete, search, start
- **VNFS servers:** create, list, get, update, delete, search, start

Everything else is dropped from the MCP. Add an `operation_id` to expose more.

Single-file change (`server.py`). Independent of #241 — the list references `add_tool_from_code` (added by #241); `include_operations` ignores ids with no matching route, so this is safe in any merge order (that op simply starts being exposed once #241 lands).

### DIFF
--- a/src/skillberry_store/fast_api/server.py
+++ b/src/skillberry_store/fast_api/server.py
@@ -1,6 +1,6 @@
 import logging
 import os
-from typing import Any, Literal
+from typing import Any, List, Literal
 
 import uvicorn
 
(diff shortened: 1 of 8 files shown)


eranra:
@PrielHaz Thanks, **I like the idea of exposing only a curated subset of operations through the Control MCP**. However, I don't think the implementation should rely on maintaining a separate list of tools alongside the FastAPI endpoints, since it will need to be updated every time a tool is added or removed.

Instead, can we label the FastAPI endpoints and use that label to determine which tools are exposed (i.e., a whitelist)?

For example, in `tools_api.py` we already use:

`openapi_extra={"x-cli-name": "get-tool"}`

to indicate that an endpoint should be exposed as `get-tool` in the CLI. I think we should align the CLI and Control MCP by using the same metadata.

One approach would be to iterate over all FastAPI endpoints, select those that define `x-cli-name`, and automatically expose them through the Control MCP. This keeps the CLI and Control MCP in sync and avoids maintaining the same configuration in multiple places.

eranra:
@PrielHaz in addtion I see that the "pull_request / pull-request (pull_request)" CI is failing and need to be fixed

PrielHaz:
Thanks @eranra — agreed on driving this off endpoint labels instead of a hand-maintained list. One caveat on reusing `x-cli-name` directly: the CLI surface is a **superset** of what should be on the agent MCP. Endpoints like `purge-all`, `backup`, `restore` carry `x-cli-name` but shouldn't be agent-callable, and `create_tool`/`add_tool` carry it too but are multipart file-uploads that *can't* be MCP tools at all (no file body in an MCP call). So selecting on `x-cli-name` would over-expose.

I went with your approach but added a dedicated marker: endpoints opt in with `openapi_extra={"x-mcp-tool": True}` right next to `x-cli-name`, and the Control MCP list is derived from the OpenAPI schema at startup (`mcp_operations_from_openapi`). Same benefit you asked for — co-located label, no parallel list that drifts — just a purpose-specific key so CLI and MCP can diverge where they must. Result: 34 operations exposed; admin/health/file-upload endpoints stay off. Added a unit test for the derivation.

eranra (approved):

PR #264, opened 2026-07-05
Fast GitHub Skill Download

erezh16:
Efficient shallow clone of full GH folder + import from local storage + prewarm cache
Improve GUI skill detect and download from Anthropic from >5 min to ~37 seconds (17 skills)

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -65,7 +65,7 @@ _session_log_handler = None
 
 
 @pytest.fixture(scope="session")
-def run_sbs():
+def run_sbs(tmp_path_factory):
     """Start the SBS server once per session in a daemon thread."""
     from skillberry_store.fast_api.server import SBS
     from skillberry_store.tests.utils import clean_test_tmp_dir, wait_until_server_ready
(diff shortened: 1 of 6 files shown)


erezh16:
It seems the security fix broke the cache, which silently failed. I marked this PR `do-not-merge` until I fix this,.

eranra:
Three main changes
Local sparse-clone cache (importer.py, ~200 lines)

On first import of a repo, does a single git clone --depth=1 --filter=blob:none --sparse into a temp cache dir
Every subsequent import from the same (owner, repo, ref) reads from disk — no network
LRU eviction keeps cache under 2 GB
Falls back silently to the old REST walk if git fails
Pre-warm (skills_api.py + AnthropicSkillImporter.tsx)

The UI now passes prewarm=true when calling detect-anthropic-skills
The server immediately spawns a background daemon thread to start the clone
By the time detect returns and the user clicks "import", the clone is already on disk
Result: ~37 seconds for 17 skills (down from >5 min)
gh-CLI keyring token fallback (endpoint_auth.py)

Old code only read ~/.config/gh/hosts.yml — missed tokens stored in the OS keyring
New code falls back to running gh auth token --hostname github.com to retrieve the token via the CLI
Also fixes a bug where override_token was ignored when no endpoint config matched
Security hardening (CodeQL)
Input from GitHub URLs is now validated against strict allowlist regexes before being used as git CLI arguments or filesystem paths — preventing CLI injection and path traversal.

UX addition
A warning now appears under the "Anonymous login" checkbox explaining GitHub's 60 req/hour rate limit for unauthenticated access.

LGTM

eranra (approved):

PR #267, opened 2026-07-07
fix: python:3.11 via GHCR mirror + bump runspace-agent to 0.4.2

PrielHaz:
## Problem
The `pull_request` CI job intermittently fails at the `pull-docker-image` step with:

```
docker pull public.ecr.aws/docker/library/python:3.11
Error response from daemon: toomanyrequests: Rate exceeded
```

AWS ECR Public throttles **anonymous** pulls per source IP; GitHub Actions shared runner IPs regularly exceed that limit. The same image is pulled at runtime by the `file_executor` sandbox.

## Changes
1. **Fix the rate limit** — pull `python:3.11` from the org GHCR mirror `ghcr.io/skillberry-ai/python:3.11` (public, multi-arch) in both places:
   - `.github/workflows/pull_request.yml` (CI pre-pull)
   - `src/skillberry_store/modules/file_executor.py` (runtime sandbox)

   GHCR public pulls aren't subject to the ECR anonymous per-IP limit, and CI authenticates with the built-in `GITHUB_TOKEN`.

2. **Bump `runspace-agent[claude]` 0.4.1 → 0.4.2** across the three dependent plugins (ask-runspace, anthropic-skill-generator, skill-optimizer). 0.4.2 builds its container image `FROM` a published multi-arch GHCR base instead of installing Python/Node/the Claude Code CLI on every build.

Closes #249

PR #268, opened 2026-07-07
chore: bump runspace-agent[claude] to 0.4.2

PrielHaz:
## What
Bump `runspace-agent[claude]` from `0.4.1` → `0.4.2` in all three plugins that depend on it:
- `plugins/skillberry-plugin-ask-runspace`
- `plugins/skillberry-plugin-anthropic-skill-generator`
- `plugins/skillberry-plugin-skill-optimizer`

## Why
runspace-agent 0.4.2 builds its container image `FROM` a published multi-arch base image (`ghcr.io/skillberry-ai/runspace-agent-base`) rather than installing Python + Node.js + the Claude Code CLI on every build. Result: `runspace-srv` image builds are fast and no longer hit apt/npm registry rate limits. No public API changes in runspace-agent.

## Note
This branch is cut from `main`, which does not yet include the CI mirror fix in #267. Its CI `pull-docker-image` step may still hit the ECR `toomanyrequests` flake until #267 merges — that's unrelated to this dependency bump.

### DIFF
--- a/plugins/skillberry-plugin-anthropic-skill-generator/pyproject.toml
+++ b/plugins/skillberry-plugin-anthropic-skill-generator/pyproject.toml
@@ -9,7 +9,7 @@ description = "Skillberry Store plugin for generating Anthropic skills from desc
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-    "runspace-agent[claude]==0.4.1",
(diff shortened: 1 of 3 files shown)


PrielHaz:
Folded into #267 (the version bump now rides along with the GHCR mirror fix, so it's a single PR to review).

PR #270, opened 2026-07-09
WIP: Support for installing with npx

erezh16:
Feature: when a skill is exported with WebDAV, it can be installed to an agent directly with npx
- **feat(anthropic): add validate_skill_slug shared naming rule**
- **feat(anthropic): validate skill slug at export, add allow_invalid_name escape**
- **feat(vnfs): add npx_compat flag to VnfsSchema with webdav-only validator**
- **feat(anthropic): emit .well-known/agent-skills layout when npx_compat=True**
- **feat(vnfs): thread npx_compat through server/manager/service; expose install_url**
- **feat(ui): npx_compat opt-in, install command card, well-known slug validation**
- **Test for npx**
- **Lint fix**


eranra:
@erezh16 thanks for suggesting this enhancment. 

I am looking at the code --- there is so much NPX noise in the code, both in the UI and in the backend ... I understand the motivation to allow installation by NPX but maybe this is not the way to go.  Can we do something that is an external converted from the common Anthropic code to the NPX code? maybe there is a python or TS module for this? we do not want all of that noise in our code unless we have to.

eranra:
@erezh16 also can you open an issue that explains what are you trying to propose here: the title "Support for installing with npx" is not infomative enought

PR #277, opened 2026-07-11
Skills import benchmark fixes

erezh16:
1. Dropped --sitemap-only - seems very limited and useless 2. When cloning skills, remove all the empty/broken repos automatically


eranra (approved):
@erezh16 can you move this from the examples folder to some other sub-folder? maybe benchmarks  ( or performance benchmanrks) sub-folder. It is not examples, so not reason for this to be in the examples sub-folder.

Issue #256, opened 2026-06-30
Demonstrate store managing 5000 skills


erezh16:
Evidence of skillberry-store robustness and scale - prove that it can manage a very large amount of skills.

Scenario:
* Single instance of skillberry-store.
* Possible - set up a constrained environment (limit CPU/memory) - e.g., using a container
* Import 5K skills from e.g., https://skills.sh
* Demonstrate quick response - UI, CLI, API, plugins

We may add this demo to the skillberry-store blog.

erezh16:
PR #278 
PR #279 
PR #280

erezh16:
Demonstrated successfully scalable UI with 3K skills. Closing with the successful merge of PR #279 which contains all the previous PRs as well: #278 #286 #287 #288 #280 #279

PR #278, opened 2026-07-11
feat: field selection via optional ?fields query param (Phase 1)

erezh16:
## Summary (Strategy A)
- Adds optional `?fields=list` / `?fields=<csv>` projection to both **list** and **search** endpoints for snippets, skills, and tools. Omitting `fields` preserves the current bare-array + full-object list shape and the legacy `{filename, similarity_score}` search shape, so CLI, SDK, and MCP callers are unaffected.
- The `list` preset drops the heavy fields the list view never needs: snippet `content`; tool `params`/`returns`/`dependencies`/`packaging_params`; skill inlined `tools`/`snippets` (UI switches to `tool_uuids`/`snippet_uuids` count on the cards + list view).
- When `fields` is set on `/search/{type}`, each match becomes a projected object with `similarity_score` merged in — saves consumers a follow-up detail fetch.
- Fixes two pre-existing cache-mutation bugs:
  - `SkillsService.list_all` was calling `populate_objects` on the dicts returned by `list_all_dicts()`, which are shared refs into `DictCache`. Now populate runs on a shallow copy.
  - All three services' `search` were assigning `similarity_score` back onto the dereferenced cache dict. Now assembled on a fresh candidate dict.
- UI list APIs (`snippetsApi.list`, `skillsApi.list`, `toolsApi.list`) hardcode `?fields=list`; all other call sites unchanged.

## Test plan
- [x] New unit tests: `test_list_projections.py`, per-service default/preset/allowlist tests for both `list_all` and `search`, plus cache-mutation regressions on both paths.
- [x] New FastAPI integration tests (`test_server.py`): default listing returns bare array with `content`; `?fields=list` drops it; `?fields=uuid,name` restricts keys; `/search/snippets` default returns legacy `{filename, similarity_score}`; `/search/snippets?fields=list` returns projected snippet + score.
- [x] Full backend suite: 1144 passed, 22 skipped, 5 pre-existing failures (confirmed by re-running on unmodified `main` — 4 in `tests/modules/test_vmcp_server.py`, 1 in `plugins/skillberry-plugin-provenance`).
- [x] UI typecheck: no new errors introduced (baseline 152 preexisting → 152 after).
- [x] Manual: browse Snippets / Skills / Tools pages in dev UI to confirm rendering; try semantic search with/without `?fields`.

### DIFF
--- a/docs/gui-performance-strategies.md
+++ b/docs/gui-performance-strategies.md
@@ -1,0 +1,399 @@
+# Strategies for Scaling the Skillberry-Store GUI to Large Object Counts
+
+Author: analysis snapshot, 2026-07-01
+Scope: `src/skillberry_store/ui` (React + PatternFly + TanStack Query + Vite) and
+its collaborating FastAPI list endpoints under `src/skillberry_store/fast_api`.
+
+---
+
+## 1. What is actually slow — root causes
+
+The symptoms (long page load, "kill this tab?" from the browser, worst on the
+Snippets page) are the product of **four independent bottlenecks that compound**:
+
+### 1.1 Server returns the full object on every list call
+
+- [`GET /snippets/`](../src/skillberry_store/fast_api/snippets_api.py#L82) →
+  [`SnippetsService.list_all`](../src/skillberry_store/services/snippets_service.py#L200) →
+  `handler.list_all_dicts()` returns **every field of every snippet, including
+  `content`** — which is the entire snippet body (may be kilobytes to hundreds of
+  kilobytes per row). With 5K+ snippets this can easily exceed 50–200 MB of
+  JSON on the wire.
+- [`GET /skills/`](../src/skillberry_store/fast_api/skills_api.py#L85) →
+  [`SkillsService.list_all`](../src/skillberry_store/services/skills_service.py#L290)
+  runs [`populate_objects`](../src/skillberry_store/services/skills_service.py#L146)
+  **for every skill**, which resolves every `tool_uuid` and `snippet_uuid`
+  reference into a **full inlined object** (including snippet `content` again).
+  For 972 skills that reference ~2K tools and thousands of snippets, this
+  materializes a huge cross-product payload — the same snippet body may appear
+  inlined in many skills.
+- [`GET /tools/`](../src/skillberry_store/fast_api/tools_api.py) also returns the
+  full tool manifest (params/returns/dependencies) for every row.
+
+### 1.2 Client fetches ALL rows on every page mount
+
+Every list page uses `useQuery({queryKey: ['<type>'], queryFn: <api>.list })`
+with no `limit`/`offset` and a 30-second `staleTime`
+(`ui/src/main.tsx:18`). The Skills page additionally pulls **all tools and all
+snippets in parallel** so the "Create Skill" modal has data ready
+([SkillsPage.tsx:103-112](../src/skillberry_store/ui/src/pages/SkillsPage.tsx#L103-L112)) —
+even when the user only wants to browse skills.
+
+### 1.3 Every row is rendered synchronously (no virtualization)
+
+The PatternFly `Table` in [`SnippetsPage.tsx`](../src/skillberry_store/ui/src/pages/SnippetsPage.tsx#L457-L542)
+and [`SkillListView.tsx`](../src/skillberry_store/ui/src/components/SkillListView.tsx#L27-L110)
+maps directly over the filtered array. React reconciles ~5K rows × 6–7 cells
+each = ~30K–40K DOM nodes on Snippets alone. Filtering, sorting, and typing
+into the search box each rerun `useMemo` over the full array and rebuild the
+whole table.
+
+### 1.4 Background poll fully invalidates the massive queries
+
+[`useChangesMonitor`](../src/skillberry_store/ui/src/hooks/useChangesMonitor.ts)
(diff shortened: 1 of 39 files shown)


eranra:
@erezh16  Does this breaks backward compatibility?

eranra:
@erezh16 can we open an issue ( or do we have an issue for this PR)?

src/skillberry_store/services/field_selection.py
```diff
@@ -0,0 +1,130 @@
...
+(snippet ``content``; tool ``params`` / ``returns`` / ``dependencies`` /
+``packaging_params``; skill inlined ``tools`` / ``snippets``). Metadata used
+by list-view UIs stays in.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Dict, Iterable, List, Optional, Set
+
+SNIPPET_LIST_FIELDS: Set[str] = {
```
eranra, line 20: Is there a way not to create hardcoded lists of fields and move that to the objects schemas (for example put some labels in the manifest to say that this is includes or not in the sub-set-of-fields) ?

src/skillberry_store/services/list_projections.py
```diff
@@ -0,0 +1,130 @@
+"""Field-projection helpers for the bulk list endpoints.
```
eranra: I am not "in-love" with the term "Field-projection" - we are talking about field-filtering and the term projection is confusin. 

I think that this is similar to other solutions like the "wide" mode in kubectl ... that you can have default, wide or you can specify list of fields ... 

```In Kubernetes, the flag -o wide (or --output=wide) is appended to kubectl get commands to display additional columns of detailed information about cluster resources. The default kubectl get view provides only a basic summary, while the wide format ```

Can we simplify to support exactly this abvstraction of default, wide or list of fields?

eranra (changes requested):
This is a very import PR for the scalability work ... I think that this is suport good progress in the correct direction. I added some comments.

src/skillberry_store/services/list_projections.py
erezh16: Projection is the mathematical term for "field filtering" in relational algebra. I suppose that's what Claude proposed this name.

erezh16:
> @erezh16 Does this breaks backward compatibility?

No. I changed the original API proposal (the one in the discussion doc that we both read last week) so that all windowing arguments (including the future ones in strategy C) are optional with default to current behavior.

erezh16:
> @erezh16 can we open an issue ( or do we have an issue for this PR)?

Issue #256

src/skillberry_store/services/list_projections.py
erezh16: Just trying to understand - are you asking to make the narrow view default and the wide view optional (opposite to the current default, which is default wide and narrow optional based on lists)?

erezh16:
BTW: the usability was tested with almost 3K skills. UI is still quite clunky. That's why I'm continuing with strategy C as well (pagination).

src/skillberry_store/services/field_selection.py
erezh16: The reason for `presets` (i.e., hard-coded lists) is to allow useful shortcuts. The basic API allows full custom `allowlists` of comma-separated field names (whitelists). Check the server FastAPI.

erezh16: TODO: add tag to field to associate with presets

src/skillberry_store/services/list_projections.py
erezh16: TODO: 
1. Default: narrow - minimal details
2. Wide - option - all the details,

erezh16: PR #286

src/skillberry_store/services/field_selection.py
erezh16: PR #286

src/skillberry_store/services/list_projections.py
erezh16: Rename from `projection` to `field selection` was added in this PR - see below

src/skillberry_store/services/snippets_service.py
aviweit, line 304: I know its not related to this PR but:
The key is always "filename", never "name". So the or m.get("name") fallback is dead code — it can never trigger.

src/skillberry_store/services/tools_service.py
aviweit, line 531: Same comment for "name" key as in snippet_service.

src/skillberry_store/tests/fast_api/test_server.py
aviweit, line 1: Skills and tools have zero HTTP-level tests for the ?fields feature.

docs/gui-performance-strategies.md
aviweit, line 1: I wonder whether it's OK to commit that with the PR? It seems to be more internal.

aviweit (commented):
@erezh16 , the changes look good to me.

I ran make test and make test-e2e as well as doing some tests with skills/tools/vmcpservers I created.

I am about to test end to end with skillberry benchmarks (tau2). If that passes, I will approve.

I will update.

Please also find several inline comments. 
Thanks.

aviweit (approved):
@erezh16 , I approve.
Thanks

erezh16:
Merged with PR #279

PR #279, opened 2026-07-11
feat: server-side pagination + /facets across all 5 list endpoints

erezh16:
## Strategy C

Stacks on top of #280  (UI fix for skills page, after strategy A). Combines Phase 2 (server-side pagination surface), Phase 3 (`/facets` endpoints + UI switch), and the vMCP/vNFS extension so all five list endpoints behave the same way.

## Summary

**All five list endpoints (``/snippets/``, ``/skills/``, ``/tools/``, ``/vmcp_servers/``, ``/vnfs_servers/``) now support the same query grammar:**

- ``fields`` — projection spec (``list`` preset / ``full`` / comma-separated allowlist).
- ``search`` — case-insensitive substring over ``name`` + ``description``.
- ``tags`` — repeatable AND-semantics tag filter (namespace tags included).
- ``state`` — exact-match lifecycle state.
- ``sort`` — ``field:direction``, default ``modified_at:desc``.
- ``limit`` + ``offset`` — bounded slice.

Response shape stays a **bare array** when neither ``limit`` nor ``offset`` is set (100% back-compat for CLI/SDK/MCP on snippets/skills/tools). Setting either flips to an envelope ``{items, total, offset, limit}``. Filter → sort → paginate runs **before** projection.

**Per-type page-bounded expensive work:**
- Skills' ``populate_objects`` (the tool/snippet fan-out) runs only on the current page.
- vMCP/vNFS runtime enrichment (``server_manager.get_server`` calls) runs only on the current page.
- Regression tests lock both in.

**Breaking wire change on ``/vmcp_servers/`` and ``/vnfs_servers/``:**
Response was previously ``{"virtual_mcp_servers": {uuid: info}}`` / ``{"virtual_nfs_servers": {uuid: info}}``. Now a bare array (or envelope when paginated), matching the other three endpoints. Rationale: with skills often exposed multiple times as vMCP/vNFS, scale between the five types is not asymmetric — consistency is the higher-value property. Pre-existing tests and ``StoreAPI.list_vmcps`` updated per user approval to match the new shape.

**Also fixes** the same cache-mutation bug in ``search`` for vMCP and vNFS that Phase 1 fixed for the other three types.

**New /facets endpoints:** ``GET /facets/{snippets,skills,tools,vmcp_servers,vnfs_servers}`` return ``{tags, namespaces, states}``. Implemented as pure ``services/facets.py::compute_facets`` (algorithm) → ``service.facets()`` (delegation) → thin FastAPI handler (data + exception translation only).

[...]

PR #280, opened 2026-07-13
perf(ui): gate SkillsPage tools/snippets preload on modal open

erezh16:
## Summary
- Adds `enabled: isCreateModalOpen` to the two `useQuery` calls in `SkillsPage.tsx` that fetch the full tools and snippets lists for the "Create Skill" modal's pickers.
- Stacked on top of #278 (phase1 list projection). Depends on that PR — please merge #278 first. The incremental change is a single commit; the rest of the diff shown here is #278.

## Why
Even with `?fields=narrow` from #278, mounting the Skills page still parses two large arrays that the user did not ask for. On a store with ~5 130 tools and ~12 608 snippets, every Skills-page mount (and every `/changes` invalidation while the page is open) re-downloads and re-parses ~17 000 objects, keeping the tab clunky when switching between list pages.

The tools and snippets data are only used when the user opens the "Create Skill" modal to pick items. Gating the two queries on `isCreateModalOpen` removes the fetches from steady browsing state; they run once the user actually opens the modal, and TanStack Query serves cached data on subsequent opens.

## Behavior
- Skills-page mount: no longer triggers `GET /tools/?fields=list` or `GET /snippets/?fields=list`.
- Opening the Create-Skill modal: same behavior as before — the two queries fire and populate the pickers.
- Removable "selected tool/snippet" chips inside the modal keep resolving via `allTools?.find(...)` / `allSnippets?.find(...)`; those lookups run inside the modal render, so the data is populated by the time chips are displayed.
- No API contract change.

## Composability with phase2 (#279)
Phase2 rewrites the *primary* skills query block but leaves the two picker preload queries at the same lines. This change modifies only those two queries, so a rebase of phase2 onto this PR should be clean. Phase2 can (as a follow-up) migrate the picker to `listPaged` with a server-side `search` query for even better UX; this PR keeps the non-paged endpoint contract to avoid depending on phase2's server work.

## Test plan
- [x] `SkillsPage.cascade-delete.test.tsx` — passes.
- [x] `npx vitest run` — no new failures introduced (pre-existing failures in `VMCPServerDetailPage.*`, `AnthropicSkillImporter.test.tsx`, `openApiGenerator.error.test.ts`, and `e2e/vmcp-server-openapi-download.spec.ts` are identical on baseline).
- [x] `tsc --noEmit` — no new type errors in `SkillsPage.tsx`.
- [ ] Manual: load Skills page with a large store, confirm no `GET /tools/` / `GET /snippets/` fires until the Create-Skill modal is opened; open the modal, confirm pickers populate normally.

### DIFF
--- a/docs/gui-performance-strategies.md
+++ b/docs/gui-performance-strategies.md
@@ -1,0 +1,399 @@
+# Strategies for Scaling the Skillberry-Store GUI to Large Object Counts
+
+Author: analysis snapshot, 2026-07-01
+Scope: `src/skillberry_store/ui` (React + PatternFly + TanStack Query + Vite) and
+its collaborating FastAPI list endpoints under `src/skillberry_store/fast_api`.
+
+---
+
+## 1. What is actually slow — root causes
+
+The symptoms (long page load, "kill this tab?" from the browser, worst on the
+Snippets page) are the product of **four independent bottlenecks that compound**:
+
+### 1.1 Server returns the full object on every list call
+
+- [`GET /snippets/`](../src/skillberry_store/fast_api/snippets_api.py#L82) →
+  [`SnippetsService.list_all`](../src/skillberry_store/services/snippets_service.py#L200) →
+  `handler.list_all_dicts()` returns **every field of every snippet, including
+  `content`** — which is the entire snippet body (may be kilobytes to hundreds of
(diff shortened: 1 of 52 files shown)


erezh16:
Merged with PR #279

PR #282, opened 2026-07-13
feat(ask-runspace): add 'dream' preset for trajectory-driven skill optimization

### LINKED ISSUE #281: ask-runspace: add a "dream" preset for trajectory-driven skill optimization
### Summary
Kagenti is adding a skill "dreaming" flow that calls `ask-runspace /run` to
improve an agent's skills from its execution trajectories. To make that a
first-class, discoverable task (and usable from the store UI directly), add a
`dream` preset to `ask-runspace` alongside the existing `optimize` preset.

### Proposal
A preset whose prompt: given an agent's recent trajectories (inlined or via an
attached trajectories MCP), identify the skills involved and improve them **in
place** via the store control MCP — update each under its existing name so the
store chains the previous version as `parent` (no `*_optimized` duplicates).
Enable the experimental agent-teams workflow, like `optimize`.

No behavioral change to the plugin itself — the request text is still the whole
prompt and the store MCP is already pre-wired; this is a starter template plus a
test.

PrielHaz:
# feat(ask-runspace): add a "dream" preset for trajectory-driven skill optimization

Closes #281

## What
Adds a first-class `dream` preset to `ask-runspace`, alongside `optimize`. Given
an agent's recent execution trajectories (inlined, or via an attached trajectories
MCP), the RunSpace agent identifies the skills involved and improves them **in
place** via the store control MCP — updating each under its existing name so the
store chains the previous version as `parent` (version history preserved; no
`*_optimized` duplicates). Enables the experimental agent-teams workflow, like
`optimize`.

This is what Kagenti's skill **"dreaming"** flow calls into (see the kagenti
`feat/skill-dreaming` PR).

## Notes
No behavioral change to the plugin — the request text is still the whole prompt
and the store control MCP is already pre-wired. This adds a starter template and
a test. Comment on the existing agent-teams test updated to reflect that both
`optimize` and `dream` enable the flag.

## Tests
`plugins/skillberry-plugin-ask-runspace/tests/test_ask_runspace_plugin.py` — added
`test_dream_preset_optimizes_skills_in_place`; full file passes (28 tests).

Assisted-By: Claude Code

## Demo (full UI walkthrough, live)

📹 **Full working demo on a real cluster:** https://drive.google.com/file/d/1x5LTkZ-_pxrb0ulSKDwWchnOJF4Ccxu9/view

Real Kagenti UI end-to-end: converse with the weather agent → click the **💤 Dream** button on the agent page → Kagenti reads the agent's new spans from **Phoenix** and sends them to a **RunSpace** session (Claude Code on the in-cluster store, via IBM LiteLLM) → the `weather-reporting` skill is **rewritten in place** (thin → transport fallback, session reuse, graceful errors, JSON interpretation, 'not a forecast' caveat + attached snippets) → **auto-sync** to `team1`. No human edits.

### DIFF
--- a/plugins/skillberry-plugin-ask-runspace/src/skillberry_plugin_ask_runspace/presets.py
+++ b/plugins/skillberry-plugin-ask-runspace/src/skillberry_plugin_ask_runspace/presets.py
@@ -20,6 +20,19 @@ PRESETS: List[Dict[str, Union[str, List[str], Dict[str, str]]]] = [
      # Optimization benefits from the multi-agent team workflow, so turn on the
      # experimental agent-teams flag for this preset (merged into the agent env).
      "env": {"CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS": "1"}},
+    {"id": "dream", "label": "Dream — improve an agent's skills from its trajectories",
+     "prompt": ("Run a skill 'dreaming' pass. You are given an agent's recent execution "
+                "trajectories (inlined in this request, or fetched via an attached "
+                "trajectories MCP). Study where the agent struggled, repeated work, or hit "
+                "errors, and identify which of its skills were involved. Then improve those "
+                "skills IN PLACE via the skillberry-store MCP: update each skill under its "
+                "existing name so the store chains the previous version as its parent "
+                "(version history is preserved) — do not create *_optimized duplicates. "
+                "Record a short rationale per change, and make NO change to a skill when the "
+                "trajectories don't justify one."),
+     "skills": [],
+     # Dreaming reviews multiple trajectories and skills; the agent-teams workflow helps.
+     "env": {"CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS": "1"}},
     {"id": "skill", "label": "Create a skill that…",
      "prompt": "Create a new skill that <describe it>, following the Anthropic skill format (a SKILL.md plus supporting files).",
      "skills": ["https://github.com/anthropics/skills/tree/main/skills/skill-creator"],
--- a/plugins/skillberry-plugin-ask-runspace/src/skillberry_plugin_ask_runspace/presets.py
+++ b/plugins/skillberry-plugin-ask-runspace/src/skillberry_plugin_ask_runspace/presets.py
@@ -24,12 +24,17 @@ PRESETS: List[Dict[str, Union[str, List[str], Dict[str, str]]]] = [
      "prompt": ("Run a skill 'dreaming' pass. You are given an agent's recent execution "
                 "trajectories (inlined in this request, or fetched via an attached "
                 "trajectories MCP). Study where the agent struggled, repeated work, or hit "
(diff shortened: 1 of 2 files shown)


erezh16 (approved):
Cool feature and demo!

This is NOT an in-place update, despite the prompt. The new skill has the same name as the old skill, but it has a UUID of its own. I verified this through the demo video.
However, it is done the way I believe is better. Skills (and all other objects) should be **immutable**. Changing skill content should always result in a new skill with new UUID. Then, you can get the latest version  using name, or a specific version using UUID.  That's how you get stability, tractability and lineage. Just like git.
@eranra FYI.

erezh16:
Not sure about that last commit. I approved the previous commit without additional changes. 
I only added as a comment that this wasn't an in-place update (new skill keeps the same UUID) but rather a regular skill evolution (new skill keeps the same name, but has its own UUID), which is how I believe things should be done. The comment was mostly for @eranra as we had a discussion about immutability. 
Why was another fix for the prompt required?

erezh16:
I think the new prompt is misleading in making the agent think that updating the skill will necessarily result in a new skill (new UUID, same name). This is NOT TRUE. `skillberry-store` clearly allows in-place updates that keep the UUID, but your agent chose to create a new skill instead, which I applauded.

Side note: I am trying to convince @eranra to drop in-place updates and move to full immutability (no in-place updates). I used your agent as a good example, but that is outside the scope of this PR.

PrielHaz:
Thanks @erezhadad — great point, and I agree.

I updated the dreaming prompt so it no longer says "in place": it now instructs the run to keep the **same skill name** (which yields a new version) and to **do whatever the trajectories call for** — revise the involved skill, add supporting snippets, or add a genuinely new skill. Updated in both places:
- kagenti backend `compose_dream_request` (the prompt the `/dream` endpoint sends)
- skillberry-store `ask-runspace` `dream` preset

Pushed to this branch. Thanks for the review!

PrielHaz:
@erezh16 What u think should be further change in the dreaming prompt?

erezh16:
> @erezh16 What u think should be further change in the dreaming prompt?

I think that the prompt should ask the agent to behave AS IF skillberry-store does not allow in-place updates, but always resolves skill names to the latest version of the skill with that name (just like git). In other words, it should always strive to create a new skill (new UUID) with the same name - just like it did in the demo video.

PrielHaz:
Fixed @erezh16 — reworked the `dream` preset prompt around your framing and made it compact.

It now tells the agent to **behave AS IF the store never allows in-place updates**, and that a skill **name always resolves to its latest version (like git)** — so to change a skill it must always create a **new skill (new UUID) under the same name**, never edit in place and never rename (no `*_optimized`).

Trimmed from ~14 lines to ~10. Updated `presets.py` + the preset test, 28 pass. (The rossoctl side's `compose_dream_request` carries the same framing so the inlined-trajectory path and the preset agree.)

PR #285, opened 2026-07-18
fix: return uuid instead of filename from search endpoints

erezh16:
## Summary

The search endpoints (`/search/skills`, `/search/tools`, `/search/snippets`, `/search/vmcp_servers`, `/search/vnfs_servers`) returned a `filename` field whose value was actually the entity's **name** — while everything the vector DB stores and the services key on is a **UUID**. That naming inconsistency has been misleading both callers and the UI, which was matching semantic-search results against entities by name via that misleading key.

This PR renames the response field to `uuid`, sets its value to the entity UUID, and cleans up the related internal terminology so the code matches what the data actually is.

### Changes

**Vector DB layer** (`vdbs/chroma.py`, `vdbs/faiss.py`, `vdbs/lancedb.py`)
- Search results now return `{"uuid": <id>, ...}` (was `"filename"`). The redundant duplicate `"id"` key was dropped.

**Services** (`skills_service`, `tools_service`, `snippets_service`, `vmcp_service`, `vnfs_service`)
- `search()` now reads `match.get("uuid")` from the VDB result and returns `[{"uuid": entity_uuid, "similarity_score": float}, ...]`.
- Renamed loop variables (`matches`, `filtered_matches`, `match`) and docstrings to reflect actual semantics.
- Kept the current FastAPI ↔ service separation intact — FastAPI stays a thin wrapper that translates arguments/return values/exceptions; only `HTTPException` remains permitted in service logic.

- This change is breaking for SDK so all consumers with issues will be redirected to use the `0.2.0` tag.

**FastAPI layer**
- Updated docstrings on all 5 `/search/*` endpoints to say "UUIDs and similarity scores".

**UI** (`types/index.ts`, `SkillsPage.tsx`, `ToolsPage.tsx`, `SnippetsPage.tsx`, `VMCPServersPage.tsx`, `VNFSServersPage.tsx`)
- `SearchResult` is now `{ uuid: string; similarity_score: number }`.
- Semantic-search filtering matches by `result.uuid === entity.uuid` (was `result.name === entity.name || result.filename === entity.name`).

**Tests**
- The 5 e2e search tests (`test_skills_api.py`, `test_tools_api.py`, `test_snippets_api.py`, `test_vmcp_api.py`, `test_vnfs_api.py`) now capture UUIDs from create responses and assert the expected UUID appears in `result.uuid`.
- `test_descriptions_sync.py` helper looks up each result's name via `service.get(r["uuid"])`.
- `test_vdb.py` checks `search_results[0]["uuid"]`.

## Test plan

- [x] Full non-e2e Python test suite: **412 passed** (the 4 `test_vmcp_server.py` failures are pre-existing on `main`, unrelated to this change — verified via `git stash`).
- [x] `grep` confirms no `filename` references remain in any search-flow file (services, VDBs, FastAPI, description module, UI pages).
- [ ] e2e tests exercised locally (require running service).
- [x] Manual UI verification: semantic search on each of the 5 pages returns the expected items.

### DIFF
--- a/src/skillberry_store/fast_api/skills_api.py
+++ b/src/skillberry_store/fast_api/skills_api.py
@@ -251,7 +251,7 @@ def register_skills_api(
             lifecycle_state: State to filter by (e.g., LifecycleState.APPROVED).
 
         Returns:
-            list: A list of matched skill names and similarity scores.
+            list: A list of matched skill UUIDs and similarity scores.
         """
         try:
             return service.search(
--- a/src/skillberry_store/fast_api/snippets_api.py
+++ b/src/skillberry_store/fast_api/snippets_api.py
@@ -230,7 +230,7 @@ def register_snippets_api(
             lifecycle_state: State to filter by (e.g., LifecycleState.APPROVED).
 
         Returns:
-            list: List of matched snippet names and similarity scores.
+            list: List of matched snippet UUIDs and similarity scores.
 
         Raises:
             HTTPException: 503 if search is not available, 500 for other errors.
--- a/src/skillberry_store/fast_api/tools_api.py
+++ b/src/skillberry_store/fast_api/tools_api.py
@@ -319,7 +319,7 @@ def register_tools_api(
             lifecycle_state: State to filter by (e.g., LifecycleState.APPROVED).
 
         Returns:
-            list: A list of matched tool names and similarity scores.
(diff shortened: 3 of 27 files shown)


erezh16:
Withdrawing this PR - it is fixed as part of the strategy A PRs

PR #286, opened 2026-07-19
Preset As Label

erezh16:
This PR continues PR #278
Field presets are defined as labels that are associated with fields for each object type (skill, tool, etc)
Created 3 basic presets: narrow, wide and full
Narrow is the basic preset
- **refactor: express field-selection presets as per-field tags**
- **refactor: three-preset field-selection model (narrow/wide/full) with flag-gated bundling**
- **change: default field-selection preset on list/search HTTP endpoints from full to narrow**
- **fix: make narrow the semantic default in parse_fields_spec, not just at HTTP layer**
- **refactor: govern "full" through FieldTags like every other preset**


erezh16:
Merged with PR #279

PR #287, opened 2026-07-19
Minimal Preset (UUID only)

erezh16:
Based on PR #286 
Added "minimal" preset with only UUID for CLI search (backward-compatible). Default is still "narrow".
- **feat: add "minimal" preset and enforce minimal ⊆ narrow ⊆ wide ⊆ full ordering**


erezh16:
Merged with PR #279

PR #288, opened 2026-07-19
Field Selection Applied to "get" Operations

erezh16:
Based on PR #287
Extended the field selection and presets to "get" operations (retrieving a single object) across all object types
Now all retrieval operations obey the field selection, with default preset being "narrow". In addition there are "wide" and "full". Similar to kubectl.

- **feat: extend ?fields field selection to per-object get endpoints**


erezh16:
Merged with PR #279

PR #293, opened 2026-07-22
ci push fix

erezh16:
- **Added standard setup for multi-platform buildx (QEMU + buildx)**
- **Replaced old docker-push target with docker-build, which also handles push with DBT=registry**
- **File change check updated to use find and avoid too big shell commands**


erezh16:
I expunged the SDK update

PR #296, opened 2026-07-27
Access Control

erezh16:
# Access Control — Summary

skillberry-store ships with **Kubernetes-style RBAC** (roles + role bindings) applied by a FastAPI middleware. See [access-control.md](access-control.md) for the full design.

## Modes

Set via `mode:` in [access_control_config.yaml](../../access_control_config.yaml) (overridable with `SBS_ACCESS_CONTROL_CONFIG`).

| Mode | Behavior |
|---|---|
| `disabled` *(default)* | No middleware installed. All endpoints reachable — backward compatible. |
| `standalone` | Bundled username/password IdP mints opaque session tokens; every request is authenticated and authorized. |
| `delegated` *(reserved)* | Header-based trust from an upstream gateway. Rejected at startup in step 1. |

## Request flow (`standalone`)

```
Request → Allow-list check → Bearer → SessionStore → Subject
                                                       │
                                             Route → (resource, verb)
                                                       │
                             PDP: any role bound to Subject that grants it?
                              ├─ yes → handler runs (subject on request.state)
                              └─ no  → 403 (401 if identity missing)
```

- **Resource** comes from the endpoint's OpenAPI tag (`skills`, `tools`, `snippets`, `vmcp_servers`, `vnfs_servers`, `admin`, `plugins`) — overridable with `openapi_extra={"x-rbac-resource": "..."}`.
- **Verb** derives from method + path: `list`/`get`/`search`/`create`/`update`/`delete`/`execute`/`admin`. Import ≡ `create`, export ≡ `get`. Overridable with `x-rbac-verb`.

## Config shape

```yaml
mode: standalone
standalone:
  session_ttl_seconds: 43200        # 12h; env override: SBS_SESSION_TTL
  users:
    - username: alice
      password_hash: "$2b$12$..."   # produced by scripts/hash_password.py
      groups: [team-blue]
roles:
  - name: reader
    rules: [{resources: [skills, tools, snippets, ...], verbs: [list, get, search]}]
bindings:
  - name: alice-reads
    subjects: [{kind: tenant, name: alice}]
    roles: [reader]
```

Built-in roles: `reader`, `content-author`, `tool-runner`, `server-operator`, `admin` (see [access_control_config.yaml](../../access_control_config.yaml)).

## Endpoints

- `POST /auth/login` — `{username, password}` → `{token, expires_at, tenant_id}`. Unauth-listed.
- `POST /auth/logout` — revokes the bearer token. Idempotent.
- `GET /auth/whoami` — returns `{tenant_id, groups, roles}` for the current token.

Always-open regardless of mode: `/health`, `/health/ready`, `/admin/metrics`, `/docs`, `/openapi.json`, `/control_sse*`, plus the three `/auth/*` endpoints.

## Clients

- **CLI** — `sbs login` prompts for credentials, writes the bearer into `~/.config/restish/apis.json` (`chmod 0600`); restish auto-injects it thereafter. `sbs logout`, `sbs whoami`. `SBS_TOKEN` env var overrides for CI.
- **UI** — In `standalone` mode a `/login` page gates the app. `vite.config.ts` inlines the mode at start-up; `AuthContext` persists the token in `sessionStorage` and installs a `window.fetch` interceptor that adds `Authorization: Bearer` and redirects to `/login` on any 401.
- **Control MCP** — Works transparently: FastApiMCP re-dispatches tool calls through the ASGI stack, forwarding the client's `Authorization` header. Same PEP applies.

## Operator quickstart

```bash
python scripts/hash_password.py alice        # prints a bcrypt hash
# paste hash into access_control_config.yaml under standalone.users
# set mode: standalone, add a binding
# restart the store
sbs login                                    # or use the UI /login page
```

## Deferred (see §16 of the full design)

Binding `scope:` (per-namespace / per-object filters — parsed but not enforced), long-lived API tokens, HttpOnly cookies, OIDC/JWT/mTLS providers, session persistence across restarts, UI RBAC-aware hiding.

## Commits:
- **feat(access-control): implement RBAC middleware, /auth endpoints, config**
- **test(access-control): unit + integration coverage**
- **feat(ui): standalone-mode login flow**
- **feat(cli): sbs login / logout / whoami subcommands**
- **style(access-control): black formatting**


eranra:
@erezh16 can you check what exists inside Kagenti (rossoctl) and make sure that what we have is aligned

erezh16:
> @erezh16 can you check what exists inside Kagenti (rossoctl) and make sure that what we have is aligned

Done. I built an image, pushed under `latest` and ran in rossoctl. As a test, I used store UI to import 17 Anthropic skills, then saw them in rosso, then created an agent with docx skill

eranra (commented):
Can we remove all of the noise " The version of the OpenAPI document: g87d2024" in all the files ... it will continue to make noise all over also in the future

eranra (commented):
I was expecting that you will use the common FastAPI's built-in authentication/autorization ? is there a strong reason why we need to reinvent the wheel here and reimplement all of that from scratch ... it is a lot of code to support and add to the project, do we really need that? is that connected to rossoctl and how the connection is made with rossoctl ?

src/skillberry_store/access_control/mapper.py
```diff
@@ -0,0 +1,111 @@
...
+        return str(extra["x-rbac-verb"])
+    method = method.upper()
+    tags = set(route.tags or [])
+
+    if "admin" in tags:
+        admin_super_paths = ("/admin/backup", "/admin/restore", "/admin/purge-all")
+        if any(path.rstrip("/").startswith(p) for p in admin_super_paths):
+            return "admin"
+
+    if method == "GET":
```
eranra: I really don't like this ... we are enforcing the autentication by parsing the URLs?

eranra:
**Recommendation: move to FastAPI's built-in auth (`HTTPBearer` + a `require_roles`-style dependency) rather than the `BaseHTTPMiddleware` PEP.**

The deciding factor is rossoctl. Its backend ([`rossoctl/backend/app/core/auth.py`](https://github.com/rossoctl/rossoctl/blob/main/rossoctl/backend/app/core/auth.py)) already does exactly this: `security = HTTPBearer(auto_error=False)`, `dependencies=[Depends(require_roles(ROLE_OPERATOR))]`, Keycloak RS256 JWTs validated via JWKS, hierarchical `rossoctl-viewer/-operator/-admin` roles.

That means the identity model here — not the middleware — is the real issue. rossoctl callers arrive holding a Keycloak JWT; `StandaloneIdentityProvider` looks it up in its in-memory session dict, misses, and returns `401 invalid_or_expired_token`. So `sessions.py`, `/auth/login`, `hash_password.py`, the YAML bcrypt user list, and the UI login page are all work that gets deleted the day this ships as a rossoctl service. `delegated` mode is the integration point and it's currently rejected at startup.

Secondary concern, independent of rossoctl: `src/skillberry_store/access_control/mapper.py` *infers* permissions from URL substrings (`"export-" in path`, `endswith("/execute")`, `"{" in route.path`) and defaults untagged routes to `plugins`. A misnamed endpoint silently gets the wrong verb — no error, just a wrong allow or deny.

Asks, in priority order:

1. **Sketch the Keycloak/`delegated` provider before merging** — it decides whether the login stack is step-1 scaffolding or permanent parallel infrastructure.
2. **Adopt `HTTPBearer` + `require_roles`**, largely liftable from rossoctl's `auth.py`. Keep enforce-by-default via app-level `dependencies` or a startup check that refuses to boot on any unmarked route.
3. **Drop the URL heuristics** either way: require explicit `x-rbac-resource`/`x-rbac-verb`, remove the `plugins` default.
4. **Blocker:** the per-tenant MCP mounts iterate `acl_cfg.users` at boot. With Keycloak there is no static user list — that needs rethinking regardless.

Worth keeping: the `config.py`/`pdp.py` RBAC shape, and `disabled` mode — though prefer rossoctl's approach of returning a mock admin from the dependency over skipping enforcement, so handlers never have to check whether a subject exists.

src/skillberry_store/access_control/mapper.py
erezh16: I believe I have the answer for that. ACL maps roles to verbs, but in the case of admin, for example, there is no specific CRUD verb that clusters all the admin operations. So we add a mapping of verbs to routes. 
HOWEVER, I will verify this with Claude.

erezh16:
> **Recommendation: move to FastAPI's built-in auth (`HTTPBearer` + a `require_roles`-style dependency) rather than the `BaseHTTPMiddleware` PEP.**
> 
> The deciding factor is rossoctl. Its backend ([`rossoctl/backend/app/core/auth.py`](https://github.com/rossoctl/rossoctl/blob/main/rossoctl/backend/app/core/auth.py)) already does exactly this: `security = HTTPBearer(auto_error=False)`, `dependencies=[Depends(require_roles(ROLE_OPERATOR))]`, Keycloak RS256 JWTs validated via JWKS, hierarchical `rossoctl-viewer/-operator/-admin` roles.
> 
> That means the identity model here — not the middleware — is the real issue. rossoctl callers arrive holding a Keycloak JWT; `StandaloneIdentityProvider` looks it up in its in-memory session dict, misses, and returns `401 invalid_or_expired_token`. So `sessions.py`, `/auth/login`, `hash_password.py`, the YAML bcrypt user list, and the UI login page are all work that gets deleted the day this ships as a rossoctl service. `delegated` mode is the integration point and it's currently rejected at startup.
> 
> Secondary concern, independent of rossoctl: `src/skillberry_store/access_control/mapper.py` _infers_ permissions from URL substrings (`"export-" in path`, `endswith("/execute")`, `"{" in route.path`) and defaults untagged routes to `plugins`. A misnamed endpoint silently gets the wrong verb — no error, just a wrong allow or deny.
> 
> Asks, in priority order:
> 
>     1. **Sketch the Keycloak/`delegated` provider before merging** — it decides whether the login stack is step-1 scaffolding or permanent parallel infrastructure.
> 
>     2. **Adopt `HTTPBearer` + `require_roles`**, largely liftable from rossoctl's `auth.py`. Keep enforce-by-default via app-level `dependencies` or a startup check that refuses to boot on any unmarked route.
> 
>     3. **Drop the URL heuristics** either way: require explicit `x-rbac-resource`/`x-rbac-verb`, remove the `plugins` default.
> 
>     4. **Blocker:** the per-tenant MCP mounts iterate `acl_cfg.users` at boot. With Keycloak there is no static user list — that needs rethinking regardless.
> 
> 
> Worth keeping: the `config.py`/`pdp.py` RBAC shape, and `disabled` mode — though prefer rossoctl's approach of returning a mock admin from the dependency over skipping enforcement, so handlers never have to check whether a subject exists.

This is more than I planned to do. My intent was to use skillberry-store in "disabled" ACL mode (NO authentication/authorization) with rossoctl, so that it becomes a drop-in replacement to the previous skillberry-store version. This is what we discussed on Monday, and this is how I tested and completed the PR.

In the ACL design (see `docs/design/access-control.md`), I also included "delegate" mode, in which rossoctl does auth/auth and skillberry-store integrates with it. **However, I did not implement it yet - on purpose**. The reason: in the current integration it has no clear value. 
It will provide good value if I modify the client code you added to rossoctl, for example, so that skillberry-store provides to each rossoctl tenant only the skills visible to that tenant, which it currently does not. Your answer from Monday about changing rossoctl is "don't change the tree". 

Bottom line: if you want to add the security delegation that you propose above and which I already had in design in "delegate" mode, I can do so in a separate foillow-up PR, so that this PR does not grow too big.

erezh16:
Here is a report that summarizes how rossoctl currently interacts with skillberry-store. It's very naive and inefficient - basically, rossoctl reads all the skills from skillberry-store, then manages it internally completely separately and handles multi-tenancy on its own. If you want that to change, we need to change this interaction also from rossoctl side, but as I said above - probably better in a separate PR.
[rossoctl-integration.md](https://github.com/user-attachments/files/30976018/rossoctl-integration.md)

erezh16:
> I was expecting that you will use the common FastAPI's built-in authentication/autorization ? is there a strong reason why we need to reinvent the wheel here and reimplement all of that from scratch ... it is a lot of code to support and add to the project, do we really need that? is that connected to rossoctl and how the connection is made with rossoctl ?

This is interesting. I instructed Claude during design to minimize the code change. As a result, he chose to use Starlette features instead of FastAPI features. Since both are part of the routing in skillberry-store, it was ok. Using FastAPI requires adding more labels to endpoints (so every change multiplied by ~50 endpoints), he chose to implement central mechanisms instead. The lesson for me is that next time, it's not just "minimize code change", it's "maximize usage of features of the existing facilities" first and then "minimize code change" as second rule.

Anyway, I think I can change the code to be much more FastAPI compatibile and keep the feature set.

src/skillberry_store/access_control/mapper.py
erezh16: Seems I was correct.The central mapping code was used to assign  resources and verbs to all endpoints even when it cannot be derived directly from the route. However, after applying the "classic" FastAPI RBAC support, this is replaced with per-endpoint label;.

erezh16:
I made the code fully aligned with classic FastAPI RBAC support using endpoint labeling. However, as Claude predicted, this resulted in more files being changed since the endpoint were added RBAC labels. Now there is no need for central mapping at all so it is removed. Also, FastAPI handles bearer token so we have everything ready for the future `delegated` mode which will be added in a separate PR.

erezh16:
Tomorrow I will check the new version with rossoctl

erezh16:
This PR is going to be closed soon. I broke it down to 6 smaller PRs.

PR #305, opened 2026-08-18
OpenShift-compliant image hardening

erezh16:
## Summary

Harden the service image so it runs unchanged under OpenShift's restricted
Security Context Constraints while remaining fully compatible with plain
Kubernetes and Docker.

## What changed

- `Dockerfile` — adjustments needed to satisfy OpenShift's default SCC:
  - Ensure a non-root numeric UID in the image metadata
  - Group-writable paths where the container needs to persist runtime state
    (OpenShift assigns an arbitrary UID at runtime, all inside the root group)
  - Keep the image runnable as-is on Docker / Kubernetes (no runtime flags
    required for the previous targets)

## Test plan

- [ ] `make docker-build && make docker-run` on Docker (behavior unchanged)
- [ ] Deploy to a Kubernetes namespace (behavior unchanged)
- [ ] Deploy to an OpenShift project with the default `restricted` SCC and
      confirm the pod starts, serves traffic, and can write to its runtime
      paths


erezh16:
I just found an issue with the permissions so `do-not-merge`

erezh16:
Manual tests pass: make run, docker-run, rossoctl, openshift. 
Removing `do-not-merge`

PR #308, opened 2026-08-25
Memory Scale Down

erezh16:
This PR is a set of memory reduction operations that affect the overall memory (RSS) size of the container
Baseline (no skills): 785MB-812MB
Final (no skills): 232MB-239MB - over 70% reduction
No performance impact designed, and none observed in UI play
Bonus: UI is served on the same port as API/CLI/MCP so 80% of services are availale on a single port. This resulted from moving the UI files to be served from FastAPI after being built by vite.

- **perf(memory): embed via fastembed/onnx instead of sentence-transformers/torch**
- **perf(memory): import vector-DB backends on demand**
- **perf(memory): cap glibc malloc arenas at 2 in the runtime image**
- **perf(memory): import OpenTelemetry only when tracing is enabled**
- **perf(memory): ship a core-only image, plugins opt-in**
- **fix(docker): align runtime-stage PLUGIN_EXTRAS default with builder stage**
- **perf(memory): move dev/build/test tools out of runtime dependencies**
- **memory: serve pre-built UI via FastAPI StaticFiles (#6)**
- **fix: set Vite base='/ui/' so StaticFiles asset paths resolve correctly**
- **fix: remove /api prefix from AnthropicSkillImporter fetch calls (#6 follow-up)**
- **fix: SPA routing under /ui basename (#6 follow-up)**
- **fix(ui): stop browsers serving a stale SPA bundle from cache**
- **fix: remove /api prefix from all remaining UI fetch calls (#6 follow-up)**


eranra:
## Post-merge review of #308 — memory wins, per-change impact, and behavioural fallout

I read all 13 commits individually (`a82288a` → `8a241bd`), plus the surrounding build/CI machinery, and reproduced the new `/ui` handler locally to check its behaviour empirically. Inline comments are attached to the specific lines where a change is needed; the items with no anchorable diff line are collected at the bottom of this comment.

**Bottom line:** the memory work is careful and the measurements are honest (the malloc-arena commit actively argues *down* the usual inflated claim for that knob). Changes #1–#4, #6, #7 I'd take as-is. All the behavioural risk is concentrated in the two changes that moved deployment boundaries: **#5** (the `-full` variant CI never builds) and **#8 / #9–13** (four missed `/api` call sites, and an entirely untested UI-serving path).

---

### Baseline → final, per improvement

Numbers are the ones each commit measured; I did not re-measure RSS.

| # | Commit | Improvement | RSS before → after | Δ | Other effect |
|---|---|---|---|---|---|
| 1 | `8e6d26c` | fastembed/onnx replaces sentence-transformers + torch | 812 → 623 MB | **−189** | image 7.23 → 4.97 GB |
| 2 | `cf36ba9` | vector-DB backends imported on demand | 623 → 533 MB | **−90** | chroma/lancedb/pyarrow/pandas off startup path |
| 3 | `8250638` | `MALLOC_ARENA_MAX=2` | 527.6 → 515.9 MiB | **−11.7** | env only |
| 4 | `106d5a8` | OpenTelemetry imported only when tracing enabled | 515.7 → 502.0 MiB | **−13.7** | grpc/otel unmapped when untraced |
| 5 | `4d8c664` | core-only image, plugins opt-in | 505.3 → 307.0 MiB | **−198** (largest) | image 4.97 → 4.89 GB; **BREAKING** |
| 6 | `4e12794` | align runtime-stage `PLUGIN_EXTRAS` with builder | — | **0** | enabler: without it #5's win is voided |
| 7 | `5d03261` | dev/build/test tools out of runtime deps | ~307 → ~300 MiB | **−7** | main win is disk (jdk4py ≈ 180 MB) |
| 8 | `6f57287` | UI served by FastAPI StaticFiles, node process gone | ~300 → ~239 MiB | **−61** | UI moves to the API port |
| 9–13 | `4677832`…`8a241bd` | vite base, `/api` prefix removal, SPA basename, cache headers | 239 → 232–239 MiB | **0** | all corrective, no memory effect |

Two notes on the arithmetic. The chain does not join perfectly (533 vs 527.6, 502.0 vs 505.3) because batches were measured across rebases — read them as independent A/B pairs, not a ledger. They do reconcile: the deltas sum to −571, and 812 − 571 = 241 ≈ the 239 final, so the "over 70%" headline holds.

Ranked by value: **#5 (198 MB) ≈ #1 (189 MB) ≫ #2 (90 MB) > #8 (61 MB) ≫ #4 ≈ #3 ≈ #7.** The last three together are ~32 MB — about 6% of the total — and they are also the three lowest-risk changes in the set.

---

### Degradation / behaviour-change assessment, per improvement

| # | Functional risk | Verdict |
|---|---|---|
| 1 | Deleted `modules/description_vector_index.py` + its 140-line test | **No loss.** Confirmed at `a82288a` that nothing but its own test referenced it — it was a second, divergent fastembed path (fastembed's *default* model, not MiniLM). The index-compatibility claim is sound: MiniLM-L6-v2's ST config includes a Normalize module, so both sides emit L2-normalised vectors. One residual — see the inline comment on `vector_db_interface.py`. |
| 2 | Backend import errors move from startup to first use | **Minor change.** `SBS_VDB=chroma` on a core-only image now fails at the first embedding call rather than at boot. Both input forms still resolve (str-Enum equality), and the `ValueError` message is unchanged. |
| 3 | — | **None.** ENV line only. |
| 4 | Guard logic untouched, only import location moved | **None.** Also removes a genuinely unused `prometheus_client` import. |
| 5 | `:latest` no longer contains any plugin | **BREAKING, as labelled** — plus a gap: the `-full` variant is never built by CI. See the inline comment on `.mk/dev.mk`. |
| 6 | — | **None.** It repairs a build/runtime `ODEPS` mismatch. |
| 7 | Tools moved out of runtime deps | **One real break:** `openapi-generator-cli`, `openapi-python-client` and `toml-cli` moved to the new `[build]` extra, but `generate-sdk` installs core-only. See the inline comment on `pyproject.toml`. `make test` is fine — `test` = `dev,pytest,plugins-all`. |
| 8 | UI moves in-process | **Several behaviour changes:** `ENABLE_UI` is now dead, the UI is unauthenticated on the API port, UI port 8002 is dead, and there is zero test coverage. See inline comments on `main.py` and `server.py`. |
| 9–13 | `/api` prefix removal | **Regression for plugin-declared endpoints** — four raw-fetch call sites were missed. See the inline comment on `PluginActionForm.tsx`. |

**Things that are genuinely closed, for the record:** the path-traversal hole introduced by the hand-rolled `/ui/{path:path}` handler in `cd3ba00` was correctly fixed by the `is_relative_to(ui_root)` guard in `81cc4e9` — I verified both the raw (`/ui/../secret`) and percent-encoded (`/ui/%2e%2e%2fsecret`) forms are blocked. And registering `atexit`/signal handlers unconditionally in `main.py` is an accidental *improvement*: VMCP cleanup now runs on SIGTERM in configurations where it previously did not.

---

### Items with no anchorable diff line

**F. The entire `/ui` surface has zero automated coverage, and CI never exercises it.**
No test in the tree references `/ui`, `StaticFiles`, or the SPA fallback. `src/skillberry_store/ui/.gitignore:11` ignores `dist`, and neither `make test` nor `make test-e2e` runs `make ui-build` — so in CI `ui_dist.exists()` is always false and the mount, the root redirect, the fallback, the cache headers and the traversal guard are all dead code during the test run. On a developer machine that *has* run `make ui-build`, the app registers extra routes and goes through the RBAC startup audit differently.

That divergence is exactly why `81cc4e9` had to retrofit the `.standalone` / `.disabled` config templates — copying either over the active config hard-failed startup once the bundle existed. This is a five-fix-commit feature with no regression net under it. A handful of e2e tests that build the bundle first would have caught the basename bug, the asset-path bug, the SPA-404 bug and the stale-cache bug before they each needed their own follow-up commit.

**Also:** `conftest.py:106` still sets `ENABLE_UI=false`, which is now a no-op.

**C-adjacent. The push job has not actually published the new image yet.**
The push run on `bd471a6` ([32854430424](https://github.com/skillberry-ai/skillberry-store/actions/runs/32854430424)) died at `.stamps/ssh-agent.env` Error 1, before `docker-build` reached the docker invocation and long before `update-sdk`. This is pre-existing — `a82288a` failed the same way — but it has two consequences worth knowing: the registry `:latest` has **not** flipped to core-only yet, and the `generate-sdk` breakage described in the `pyproject.toml` comment below is currently masked rather than absent.

**H. Housekeeping.**
- `.mk/local.mk:13` still declares `SERVICE_PORTS := 8000 8002`, so the now-dead UI port is still `EXPOSE`d. Whoever owns the k8s manifests needs probes/ingress moved to `8000/ui`.
- Docs still describe the old UI model: `README.md:181` and `site/getting-started.html:131` advertise `ENABLE_UI=false`, `docs/config-env-vars.md:10` documents it as supported, and `README.md:181` still names port 3000.
- The squash-merge collapsed all 13 commit messages, so the measurement data and the `BREAKING:` note now live only on this PR page. There is no CHANGELOG in the repo for the "call out in the release notes" TODO from `4d8c664` to land in.

.mk/dev.mk
```diff
@@ -43,3 +43,15 @@ ui-clean: ## Remove UI build artifacts and node_modules
...
 
+
+##@ Docker image variants
+
+# The default image is core-only (Dockerfile sets ARG PLUGIN_EXTRAS= empty).
+# This builds the companion variant carrying every bundled plugin, tagged
+# :<version>-full / :latest-full so both can coexist in the registry. Run it
+# with `make docker-run IMAGE_TAG_SUFFIX=-full`.
+.PHONY: docker-build-full
+docker-build-full: ## Build the all-plugins image variant (tagged -full)
```
eranra, line 54: **A. `docker-build-full` exists but nothing calls it, so `:latest-full` will never be published.**

`skillberry-common/.mk/ci.mk:22` is the only place CI builds an image, and it runs exactly:

```make
VERSION=$(VERSION) DBT=registry make docker-build
```

That now produces the core-only variant. `4d8c664`'s message anticipated this precise failure — *"otherwise CI would silently start publishing slim-only, with no way to publish the full variant"* — and the `IMAGE_TAG_SUFFIX` / `EXTRA_BUILD_ARGS` hooks are the right mechanism, but the target they enable is never invoked. So the BREAKING note's own remediation (*"Deployments relying on the bundled plugins must switch to `:latest-full`"*) points at a tag that does not exist in the registry and never will.

**Fix:** add a step to `ci-push`, e.g.

```make
@echo "|||====> Executing make docker-build-full"
VERSION=$(VERSION) DBT=registry make docker-build-full
```

Minor, same file/mechanism: `docker-rmi` does `rm -f .stamps/docker-build*` and now `rm -f .stamps/docker-get*`, so removing the core image also invalidates the `-full` variant's stamps and forces a needless rebuild. Cosmetic, but worth a glob narrow while you're in here.

src/skillberry_store/ui/src/components/PluginActionForm.tsx
```diff
@@ -24,6 +24,16 @@ import type { PluginAction, PluginActionResult } from '@/types';
 // A file picked from a dropped/selected folder, with its path relative to that folder.
 type PickedFile = { file: File; path: string };
 
+// Normalise a plugin-supplied endpoint URL by stripping a leading "/api"
+// prefix if present. Plugin authors write "/api/plugins/..." because the
+// Vite proxy used to strip it before forwarding to FastAPI; now that the
+// UI is served in-process by FastAPI, no proxy exists and the routes are
+// accessed directly (e.g. "/plugins/..."). Keeping this normaliser here
+// means existing plugins do not need to be updated.
+function normalizeEndpoint(url: string): string {
```
eranra, line 33: **B. `normalizeEndpoint` is the right fix, but it covers 3 of 4 call sites in this file and 0 of 6 in three other components.** This is the one outright functional regression in the PR.

Before #6 the Vite proxy stripped `/api` via `rewrite: (path) => path.replace(/^\/api/, '')`, so plugin-declared endpoints written as `/api/plugins/...` worked in both `vite dev` and `vite preview`. With FastAPI serving the bundle directly there is no proxy, so every *raw* fetch of a plugin-supplied URL now 404s.

**Missed in this file:**
- `PluginActionForm.tsx:139` — the `x-options-from` dynamic-dropdown fetch (`const url = interpolateUrl(schema['x-options-from'], currentFormData)` at :133, fetched unnormalised at :139). Breaks `skillberry-plugin-ask-runspace`'s `"x-options-from": "/api/plugins/ask-runspace/presets"` → the dropdown silently renders empty.

**Missed entirely (files not touched by this PR):**
- `PluginNotifications.tsx:56` — `fetch(notificationsConfig.poll_endpoint)`
- `PluginNotifications.tsx:67` — `fetch(action.endpoint.replace('{uuid}', itemUuid))`
  → breaks `skillberry-plugin-dedupe`'s `poll_endpoint: "/api/plugins/dedupe/decisions"` and its `.../{uuid}/keep` + `.../{uuid}/delete` actions. The dedupe notification banner never appears and the keep/delete buttons never fire.
- `CatalogImportView.tsx:71` — `fetch(config.search_endpoint, ...)`
- `CatalogImportView.tsx:96` — `fetch(config.detail_endpoint.replace('{id}', ...))`
- `CatalogImportView.tsx:119` — `fetch(config.import_endpoint, ...)`
  → breaks `skillberry-plugin-skillssh-importer`'s `search_endpoint` / `import_endpoint`, i.e. the whole catalog-import flow.

**Why the suite stays green:** the affected components' unit tests still assert the `/api/` strings (`PluginNotifications.test.tsx:87` expects `'/api/plugins/'`, `CatalogImportView.test.tsx:15-17`, `PluginActionForm.test.tsx:344,363`), so the tests pass against endpoints that no longer resolve at runtime.

**Suggested fix:** hoist `normalizeEndpoint` out of this component into a shared util (e.g. `src/utils/endpoints.ts`) and apply it at all seven raw-fetch sites. Keeping the normaliser — rather than rewriting the plugins — is the right call as the comment says, since it keeps third-party plugins working; it just needs to be applied everywhere plugin-supplied URLs are fetched.

pyproject.toml
```diff
@@ -123,33 +93,57 @@ dependencies = [
...
+  "mypy-extensions>=0.4.4",
+  "pycodestyle>=2.7.0",
+  "pyflakes>=2.3.1",
+  "ruff>=0.9.9",
+  "skillberry-store-sdk",
+]
+
+# Build/codegen tools: Rust build tools, SDK generation (includes jdk4py ~180 MB).
+# Install with: make install-requirements ODEPS=build
+build = [
```
eranra, line 127: **C. `generate-sdk` needs this new `[build]` extra but only ever installs core deps.**

`openapi-generator-cli`, `openapi-python-client` and `toml-cli` moved here out of `dependencies`, which is correct for image size. But in `skillberry-common/.mk/dev.mk`:

```make
generate-sdk: install-requirements # Generate SDK
	...
	@openapi-generator-cli generate -i $(OPEN_API_SPEC_URL)/openapi.json ...
	...
	@toml set --to-array --toml-path $(PYTHON_SDK_DIR)/pyproject.toml ...
```

`install-requirements` with an unset `ODEPS` resolves to `uv pip install -e .` — core only. Nothing earlier in the pipeline installs `[build]` either: `ci-pull-request` runs `make test`, which is `ODEPS=test` → `dev,pytest,plugins-all`, and neither `dev` nor `pytest` carries the codegen tools. So `make update-sdk` (called by `ci-push`) will fail with `openapi-generator-cli: command not found`, then `toml: command not found`.

This is currently **masked, not absent**: the push job dies earlier at `.stamps/ssh-agent.env` (pre-existing, `a82288a` failed identically), so `update-sdk` never runs. It will surface the moment the ssh-agent step is fixed.

**Fix:** make the dependency explicit, e.g.

```make
generate-sdk: ## Generate SDK
	@$(MAKE) install-requirements ODEPS=build
```

The rest of this reorganisation checks out — `test = ["skillberry-store[dev,pytest,plugins-all]"]` keeps `make test` whole, and the builder stage's `ODEPS=${PLUGIN_EXTRAS}` correctly keeps all of it out of the runtime venv.

src/skillberry_store/main.py
```diff
@@ -41,37 +41,34 @@ def main():
...
+    # set ENABLE_UI_SUBPROCESS=true to restore the old Vite preview behaviour.
     ui_started = False
     ui_manager = None
-
-    # Check if UI should be enabled (default: True, can be disabled via env var)
-    enable_ui = os.getenv("ENABLE_UI", "true").lower() in ("true", "1", "yes")
-    
-    if enable_ui:
-        # Initialize and start UI manager with configured port
+    if os.getenv("ENABLE_UI_SUBPROCESS", "false").lower() in ("true", "1", "yes"):
```
eranra, line 61: **D. This makes `ENABLE_UI` dead, and there is now no supported way to run the store API-only.**

`main()` reads only `ENABLE_UI_SUBPROCESS` now; `ENABLE_UI` is no longer consulted anywhere. But it is still advertised as a supported switch in four places:

- `README.md:97` — `ENABLE_UI=false make run`
- `README.md:181` — *"The Web UI starts automatically on port `3000`. To disable it, set `ENABLE_UI=false`"* (both the port and the switch are now wrong)
- `docs/config-env-vars.md:10` — *"If False - disable the UI and run only the backend"*
- `site/getting-started.html:90,131`
- `conftest.py:106` still sets `os.environ["ENABLE_UI"] = "false"`, which is now a no-op

Setting `ENABLE_UI=false` today silently still serves the UI. Combined with the new `GET /ui*` / `HEAD /ui*` / `GET /` entries on the unauthenticated allow-list, that means the bundle — including sourcemaps, since `vite.config.ts` keeps `sourcemap: true` — is served unauthenticated on the *same port as the API*, and can no longer be firewalled separately the way the old port-8002 `vite preview` could.

Two decisions to make explicitly rather than by omission:
1. If API-only is still a supported mode, wire `ENABLE_UI=false` to skip the `/ui` mount and the root redirect in `server.py`.
2. If it is not, delete the switch from all four docs and from `conftest.py`, and say so in the release notes.

**Credit where due:** moving `atexit.register(cleanup_all)` and the signal handlers out of the `if enable_ui:` block is an accidental *fix* — previously, running with `ENABLE_UI=false` registered no cleanup at all, so VMCP servers leaked on SIGTERM. Worth keeping regardless of which way (1)/(2) goes.

src/skillberry_store/fast_api/server.py
```diff
@@ -331,8 +332,83 @@ def configure_fastapi(self):
...
+            )
+            async def _ui_spa_fallback(path: str):
+                # Real static assets (JS, CSS, fonts, images) are served from
+                # disk. Anything else falls back to index.html so React Router
+                # can handle the route client-side (/ui/skills, /ui/tools/:uuid).
+                asset = (ui_root / path).resolve()
+                # Unlike StaticFiles, this handler joins a client-supplied path
+                # onto the bundle directory, so it has to reject traversal out
+                # of it (e.g. GET /ui/../../../etc/passwd from a raw client).
+                if asset.is_relative_to(ui_root) and asset.is_file():
```
eranra, line 389: **E. `GET /ui/index.html` still gets a one-year immutable cache — the exact bug this commit set out to fix.**

The asset branch matches `index.html` by filename, so the un-hashed entry point is served with `_ASSET_CACHE_CONTROL` whenever it is requested by its real name. I reproduced this handler verbatim (same `resolve()` / `is_relative_to` / `FileResponse` logic) behind uvicorn:

```
/ui/skills                200  no-cache, must-revalidate      ok
/ui/                      200  no-cache, must-revalidate      ok
/ui/assets/app-abc123.js  200  max-age=31536000, immutable    ok
/ui/index.html            200  max-age=31536000, immutable    <-- pinned for a year
HEAD /ui/                 200  no-cache, must-revalidate      ok
/ui/../secret.txt         -> index.html                       traversal blocked
/ui/%2e%2e%2fsecret.txt   -> index.html                       traversal blocked
```

Normal `/` → `/ui/` traffic is fine, so this will not bite in casual testing. It bites a bookmark, a doc link, or any ingress/proxy that rewrites `/ui/` → `/ui/index.html` — and the symptom is precisely the one described in this commit message: nav renders, content pane is the router 404, no rebuild ever reaches the user, and no amount of reloading fixes it.

**Fix:**

```python
if asset.is_relative_to(ui_root) and asset.is_file() and asset.name != "index.html":
```

**Also confirmed working, for the record:** the `is_relative_to(ui_root)` guard genuinely closes the traversal hole that `cd3ba00` opened (the earlier `asset = ui_dist / path; if asset.is_file()` form would have served arbitrary files, unauthenticated, given `GET /ui*` on the allow-list). Both the raw and the `%2e`-encoded forms are blocked. Good catch in this commit — `.resolve()` also handles the symlink case.

src/skillberry_store/fast_api/server.py
```diff
@@ -331,8 +332,83 @@ def configure_fastapi(self):
...
+        # checkout), the /ui mount is simply skipped — the API still starts
+        # normally, and `make ui-dev` / `make ui-build && make run` work
+        # as before.
+        ui_dist = Path(__file__).parent.parent / "ui" / "dist"
+        if ui_dist.exists():
+            # Catch-all for SPA deep-links: any /ui/<path> that is not a real
+            # static file (JS/CSS/fonts) should serve index.html so React
+            # Router can handle the route client-side. This route must be
+            # registered BEFORE the StaticFiles mount, because Starlette
+            # evaluates mounts before APIRoutes — once StaticFiles is mounted
```
eranra, line 357: **Nit + a consequence worth stating explicitly.**

The comment is inaccurate: Starlette does *not* evaluate mounts before `APIRoute`s. `Router.__call__` walks `self.routes` in registration order and takes the first match — `Mount` has no precedence. The conclusion ("register the route first") is right, but for the ordinary reason, not a special mount rule.

The consequence, which is worth a comment because it is easy to miss: because this route is registered first and `{path:path}` matches everything, the `StaticFiles` mount two lines below is now reachable only for the bare `/ui` → `/ui/` redirect. Every actual byte under `/ui/` is served by this hand-rolled handler, so StaticFiles' own hardening and `Range` handling no longer apply to the bundle — which is exactly why the traversal guard had to be added by hand. Either say so here, or consider dropping the mount and handling the bare-`/ui` redirect explicitly, so a future reader does not assume StaticFiles is still in the path.

src/skillberry_store/vdbs/vector_db_interface.py
```diff
@@ -1,22 +1,32 @@
...
 from abc import ABC, abstractmethod
 from typing import List, Dict, Any, Optional
 
-from sentence_transformers import SentenceTransformer
+from fastembed import TextEmbedding
+
+# Same 384-dim weights that ``SentenceTransformer('all-MiniLM-L6-v2')`` used,
+# served through onnxruntime instead of torch. Both produce L2-normalized
+# vectors that agree to ~1e-7, so indices built by either remain valid.
+_ENCODER_MODEL = "sentence-transformers/all-MiniLM-L6-v2"
```
eranra, line 12: **G. Two things to verify about the fastembed swap. The change itself is sound — the lock and the model-name pin are both right — these are deployment-surface questions.**

**1. Long-text divergence is the one place the 1.4e-7 equivalence claim does not hold.** sentence-transformers caps this model at `max_seq_length=256` (from its `sentence_bert_config.json`); fastembed applies the tokenizer's own limit. For any text under ~256 tokens the two agree to ~1e-7 as measured, but a description that exceeds it gets truncated at a *different* point, which changes the vector — and therefore the ranking against an existing faiss index. Worth spot-checking against the longest `description` in a real index before relying on "indices built by sentence-transformers remain valid".

**2. Cache location and the cold-start download.** Nothing in the tree sets `HF_HOME`, `TRANSFORMERS_CACHE`, or a `cache_dir=` here, and fastembed does not read `XDG_CACHE_HOME` — it defaults to its own temp-dir cache, not the `~/.cache/huggingface` path the old code (and the old `conftest` fixture) warmed. Consequences to check:
- `/health/ready` gates on `encoder_warmup`, so the ~80 MB ONNX download is on the critical path to readiness — the commit measured 11.45 s for it. Where `/tmp` is an ephemeral `emptyDir`, that is paid on *every* pod restart.
- The image is 4.89 GB and still cannot start without reaching HuggingFace. If air-gapped deployment is in scope, pre-seeding the cache at build time (or `cache_dir=` under `$APP_HOME`, which the Dockerfile already `chgrp 0 / chmod g=u`s for the OpenShift random-UID case) is a cheap follow-up.

**3. Minor:** `fastembed` is unpinned in `pyproject.toml` while the model string is pinned here. A fastembed major bump that renames or drops `sentence-transformers/all-MiniLM-L6-v2` from its supported list would fail at first embed, not at build. Worth a lower bound at least.

eranra (commented):
Line-anchored findings from the review — see the summary comment on the PR for the impact table and the per-change degradation assessment. Ordered roughly by how much I'd want each addressed.

PR #316, opened 2026-09-03
Point the simulate plugin at the external simulation-harness

bnayahu:
## What

The simulation harness has moved out of the internal repo. It now lives at
[**skillberry-ai/simulation-harness**](https://github.com/skillberry-ai/simulation-harness)
and publishes to `ghcr.io/skillberry-ai/simulation-harness` (public — no registry
login). This points `skillberry-plugin-simulate` at that image and aligns it with
the image's **actual** contract, verified by reading the harness source at v0.1.2
rather than trusting the old assumptions.

Four of those assumptions were wrong, and the plugin would have failed against the
external image:

| # | Assumption in the plugin | Reality in the external harness |
|---|---|---|
| 1 | image `simulation-harness:latest` (local build) | `ghcr.io/skillberry-ai/simulation-harness` |
| 2 | mounts at `/skills-store` (**ro**) and `/logs` | image reads `/app/skills-store` and `/app/logs`, and **writes** to the former |
| 3 | session expiry arrives as `410` on the status endpoint | never surfaces there; the harness self-heals per tool call |
| 4 | harness may be handed an unnamed simulation | name keys a generated-skill cache shared by every simulation |

The parts that still line up are worth stating, since they are what the whole
design hangs on and they needed no change: REST port `8086`,
`POST`/`GET`/`DELETE /api/v1/simulation`, and `mcp_port` as a create field. With
`mcp_port` set, `_build_mcp_url` derives the sidecar URL from the request Host —
which is exactly the `127.0.0.1:<rest_port>` the plugin dials — so it returns
`http://127.0.0.1:<mcp_port>/mcp/sse`, the harness defaults to `sse`, and the
store connects with `sse_client` (`server_utils.py:148`). The 1:1 port publish and
the tool-manifest `mcp_url` keep working untouched.

## The change

**1. Image tag — `:0.1`, deliberately not `:latest`.** Upstream tags `latest` from
every push to `main` (`type=raw,value=latest,enable={{is_default_branch}}`), so it
is the *least* stable tag despite being the name one reaches for by convention. It
has already diverged from the newest release:

```
latest  sha256:49746ffa4bef…   # main
main    sha256:49746ffa4bef…
0.1     sha256:07be14cd236f…   # v0.1.2
0.1.2   sha256:07be14cd236f…
```

`:0.1` is the newest `v0.1.x` release and the closest thing to "latest release"
that exists today. It does stall silently on the next minor, so I filed
[simulation-harness#3](https://github.com/skillberry-ai/simulation-harness/issues/3)
asking for a `release` tag; until that lands, a `v0.2.0` needs a deliberate bump
here, which the README now says.

**2. Mount points — and read-write.** The image bakes
`HARNESS_SKILLS_FOLDER=/app/skills-store` and `HARNESS_LOG_DESTINATION=/app/logs`,
so the old `/skills-store` and `/logs` binds landed nowhere and were silently
ignored. `skills-store` also has to be **rw**: the harness generates `SKILL.md`,
`schema.json`, `db.json` and `api.json` into it, so a read-only bind fails
generation outright.

**3. Removed the resubmit-on-410 path.** `SessionExpiredError` is raised only
inside the harness's own `execute_tool`, where it resets the session itself and
fails just the one call that tripped the limit
(`core/simulation_instance.py:130-146`). It never reaches the `410` handler from a
status poll, and tool calls go from the vMCP straight to the harness, so this
client could never observe it — the code and its test were unreachable. The README
claimed *the plugin* auto-resets; corrected to say the harness does, per call.

**4. Named simulations, keyed by tool surface.** The harness caches generated skill
artifacts under `<skills-store>/<name>/` and skips the slow LLM generation on a
hit. Every simulation shares one mounted skills-store, so an unnamed simulation
(name derived from `info.title`) means two skills with the same display name
collide, and re-simulating a skill whose tools changed silently reuses stale
artifacts. Each simulation is now `<skill-slug>-<tools-fingerprint[:8]>`, reusing
the `tools_fingerprint()` that already existed for drift detection. The slug
mirrors the harness's own `sanitize_skill_name` and reserves room for the suffix,
so its 64-char truncation cannot eat the fingerprint.

**Docker calls move off the event loop.** `simulate()` is async but called
`harness_manager.start()` synchronously. That was cheap against a locally-built
image; against a registry, docker-py pulls on first use — a multi-hundred-MB
download stalling the store's event loop. Both start and stop go through
`asyncio.to_thread`, which makes `teardown()` async (its only two callers await it).

**Two smaller contract fixes.** `409` (busy harness *or* `mcp_port` already bound —
plausible when a leaked container holds it) and `422` (rejected OpenAPI spec) now
get distinct, actionable messages instead of one opaque `HarnessError`. And a
`failed` status renders the harness's structured `error` object — it is
`{code, message, details}` now, where the old code would have stringified a dict
into the exception message.

**Config passthrough.** Optional `SIMULATE_LLM_PROVIDER` /
`_SKILL_GENERATION_MODEL` / `_SIMULATION_MODEL` forward as `HARNESS_LLM_*`. The
image defaults to provider `openai` with `gpt-4.1` for both models, which need not
resolve behind a gateway that namespaces model ids. Unset means "leave the image's
`harness.yaml` alone".

## Tests

Plugin suite **62 → 82**, all passing.

I checked the two most load-bearing new tests are real regressions rather than
tautologies: reverting just the mount config and just the `name=` kwarg makes
exactly those two fail and nothing else.

New coverage: the image default is the release line and never `:latest`; mounts use
the image's paths and are writable; `HARNESS_LLM_*` appear only when set;
transitional statuses (`pending` / `generating_skill` / `initializing`) keep
polling; `409` / `422` / `404` map to distinct messages; a structured `error`
object renders; a timeout reports the last status and generation phase; and five
cases pinning the simulation-name rules (changes with the tool surface,
distinguishes same-named skills, survives the 64-char cap, is a valid Agent Skills
name, falls back when the base is all punctuation).

`make lint` clean. Note it only covers `src/skillberry_store/**`, not the plugin
tree — so I also checked `black` on the plugin separately: the tree is **16
reformat / 4 clean** both before and after this PR, i.e. it was never
black-formatted and this change does not add to that. One file I touched had
regressed to non-clean; it is formatted back.

Full `make test` shows 22 failed / 146 errors, all in `src/skillberry_store/tests/**`
(e2e wanting a live service, UI-serving tests wanting a `dist` bundle that
`make test` does not build). **These are pre-existing and unrelated.** I verified
that rather than assuming it — running the failing subset in one fixed environment
with only the plugin swapped between this branch and `origin/main` gives an
identical 22 failed / 9 passed / 43 errors either way. The plugin's own 82 tests are
collected by `make test` and pass within it.

## Not covered

Everything here is unit-level. The mount paths, the cold-start pull and the cache
naming only truly prove out end-to-end, and a smoke run means a ~500MB pull plus
real LLM spend, so it has not been done. Worth one manual run before anyone relies
on this in anger: pull the image, simulate a sample skill, confirm the sim vMCP
answers a tool call.

### DIFF
--- a/plugins/skillberry-plugin-simulate/.env.example
+++ b/plugins/skillberry-plugin-simulate/.env.example
@@ -2,11 +2,21 @@
 SIMULATE_LLM_API_KEY=
 # Optional: set for Azure/OpenAI-compatible base URL
 SIMULATE_LLM_API_BASE=
-# Container image for the simulation harness
-SIMULATION_HARNESS_IMAGE=simulation-harness:latest
+# Container image for the simulation harness. Default tracks the newest v0.1.x
+# *release* of github.com/skillberry-ai/simulation-harness. Do not use :latest —
+# upstream tags that from every push to main, so it runs ahead of any release.
+SIMULATION_HARNESS_IMAGE=ghcr.io/skillberry-ai/simulation-harness:0.1
+# Optional: override the harness's own LLM settings (its image defaults to
+# provider `openai` with `gpt-4.1` for both models). Set these when
+# SIMULATE_LLM_API_BASE points at a gateway that namespaces model ids
+# differently; leave unset to keep the image's harness.yaml defaults.
+SIMULATE_LLM_PROVIDER=
+SIMULATE_LLM_SKILL_GENERATION_MODEL=
+SIMULATE_LLM_SIMULATION_MODEL=
 # Where the plugin stores its active-vMCP registry
 SIMULATE_DATA_DIR=~/.skillberry/simulate
-# Host paths mounted into the harness container
+# Host paths mounted into the harness container. skills-store is mounted
+# read-write: the harness generates (and then caches) its skill artifacts there.
 SIMULATE_SKILLS_STORE_PATH=
 SIMULATE_LOGS_PATH=
 # Seconds to wait for harness to become ready (default: 600)
--- a/plugins/skillberry-plugin-simulate/README.md
+++ b/plugins/skillberry-plugin-simulate/README.md
@@ -1,8 +1,9 @@
 # Skillberry "Simulate This" Plugin
 
 Stand up a **simulated parallel vMCP** for a skill — a second vMCP whose tools are
-backed by a containerized simulation-harness instead of the real backends — and flip a
-plugin-owned switch deciding whether consumers reach the real or the simulated vMCP.
+backed by a containerized [simulation-harness](https://github.com/skillberry-ai/simulation-harness)
+instead of the real backends — and flip a plugin-owned switch deciding whether
+consumers reach the real or the simulated vMCP.
 
 ## How it works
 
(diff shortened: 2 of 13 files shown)


bnayahu:
@skillberry-bot can you review, please?

erezh16 (approved):
LGTM

